PR #108, opened 2014-12-10
Use atomic redis SET even when cache is used with timeout and nx

23doors:
As described in: https://github.com/niwibe/django-redis/issues/107

Atomic redis set: http://redis.io/commands/SET


niwibe:
Merged with little changes that causes test breaks ;)

Thanks!

PR #121, opened 2015-03-05
Django 1.8 compatibility

yurtaev:
- [use collections.OrderedDict instead SortedDict](https://docs.djangoproject.com/en/dev/releases/1.7/#django-utils-datastructures-sorteddict)
- [use importlib instead django.utils.importlib](https://docs.djangoproject.com/en/dev/releases/1.7/#django-utils-dictconfig-django-utils-importlib)


niwibe:
thanks! ;)

PR #123, opened 2015-04-11
future 1.9 compatibility in tests.

brandonjschwartz:
lol

I kept getting a warning regarding line 25 in django-redis/utils.py; django is deprecating the importlib module in their utilities. After loading everything up, the code causing the warning was already removed. My local copy was outdated.

Regardless, I didn't discover this until running tests and changing a few import lines that will conflict with 1.9 in the future.

Decided to commit anyway; my waste of time doesn't need to be a waste for you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ dist*
 .\#*
 .tox
 dump.rdb
+.idea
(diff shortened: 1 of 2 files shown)


niwibe:
It is ok for me, it requires some changes for make tests pass but I can do it my self

PR #125, opened 2015-04-17
Adds Sentinel support

cmheisel:
* Adds previously removed SentinelClient
* Adds diagnostic logging


niwibe:
Would be awesome have this as separated package, I can not maintain this code because I'm not using sentinel. Living in a separated package and maintained by someone that uses sentiner will be a better approach in my opinion.

cmheisel:
@niwibe Cool I setup https://github.com/KabbageInc/django-redis-sentinel -- if you've got any feedback on it, I'd appreciate it!

PR #134, opened 2015-05-26
Add support for alternate serializers to django-redis.

### LINKED ISSUE #133: Add support for alternative serializers (other than pickle)
I'm integrating django-redis into an application that shares a browser session across multiple applications. This session must be readable by applications written in languages other than Python.

When integrating django-redis, I noticed that all data sent to the cache server is first serialized using pickle. While this makes sense for Python only applications, it isn't very friendly for non-Python applications.

Would you be open to adding support for alternative serializers in the `DefaultClient` class? The default would continue to be pickle, but by adding a `JSONClient` (or whatever name) the user could serialize the data to a format that is more language agnostic.

If you're open to this idea I don't mind doing the work. I'm motivated as I'd need to do this anyway for my project. Care would be taken to factor out the common client code so only the serializing functions would be different.

This would be similar to how Django's session engine has support for multiple serializers: <https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-SESSION_SERIALIZER>

Thanks.

jdufresne:
Python pickle remains the default serializer. Added a second
serializer, JSONSerializer.

Fixes #133 

All feedback welcome. Thanks.

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -2,15 +2,6 @@
 
 from __future__ import absolute_import, unicode_literals
 
-# Import the fastest implementation of
-# pickle package. This should be removed
-# when python3 come the unique supported
-# python version
-try:
-    import cPickle as pickle
-except ImportError:
-    import pickle
-
 import random
 import socket
 import warnings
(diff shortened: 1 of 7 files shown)


niwinz:
Great work! Thanks!

PR #136, opened 2015-05-30
Fixing JSONSerializer.loads()

pawook:
In Python3, redis returns value of the key as bytes. That's good if we are using PickleSerializer, but if we use JSONSerializer, we'll get TypeError, as json.loads() accepts string, not bytes. 
When we use cache as a session backend, this error may be hided.
This means, for example, that you will not be able to login to your Django app, but you will see no errors (this happened in my project).


niwinz:
Thanks!

PR #139, opened 2015-06-03
Added serializers to setup.py

uditagarwal:
This was causing the installation to not work correctly, as it never installed the serializers folder.


niwinz:
Great!

Issue #141, opened 2015-06-08
delete_pattern uses keys instead of iter_keys


lenzenmi:
Is there any reason we couldn't change this to use a loop with iter_keys instead, or better yet, a custom lua script?

If there are a lot of keys in redis (even if they don't match your pattern) redis scans through each of them to generate the list to return. This can be quite slow, which is why the iter_keys method was introduced.

I'd be willing to take a crack at it, but I wanted to see if there was a deeper reasoning behind using the keys command, such as backwards compatability.

```python
def delete_pattern(self, pattern, version=None, client=None):
    """
    Remove all keys matching pattern.
    """
    if client is None:
    client = self.get_client(write=True)
    pattern = self.make_key(pattern, version=version)
    try:
        keys = client.keys(pattern)
        if keys:
            return client.delete(*keys)
        except _main_exceptions as e:
            raise ConnectionInterrupted(connection=client, parent=e)
```

niwinz:
As far as I know, the keys is used because iter_keys is introduced later that delete_pattern...

You are right, seems that iter_keys is a better approach for delete_pattern :D

lenzenmi:
Since you responded so quick, I went ahead and made a pull request. The tests are passing for me.

PR #142, opened 2015-06-08
delete_pattern to use scan_iter under the hood

lenzenmi:
Relates to issue #141 

I also updated the readme.txt found inside the tests folder with direction I needed in order to run the tests.


niwinz:
Thanks!  I have merged it independently of tests. I will fix them later!

Thanks again!

Issue #149, opened 2015-07-16
Improve default return values in omit_exception decorator


ecmendenhall:
Thanks for all your work on this project. We're excited to move our app over from memcached! 

We use [django-cachalot](https://github.com/BertrandBordage/django-cachalot) to cache SQL queries, and I ran into an issue when I plugged this in as my cache backend: all cache operations return `None` when `IGNORE_EXCEPTIONS` is enabled in cache configuration. That works for the default caching middleware, but cachalot blew up when it called `get_many()` expecting [an empty dict](https://github.com/django/django/blob/b74b94445de4173351b8c347deac22129c549f0f/django/core/cache/backends/base.py#L135) and instead got `None`.  

It would be great if the [`omit_exception` decorator](https://github.com/niwinz/django-redis/blob/master/django_redis/cache.py#L14) were a little smarter, returning `{}` for `get_many()` and whatever default value is passed as a keyword argument for `get()`. The [dummy cache implementation](https://github.com/django/django/blob/b74b94445de4173351b8c347deac22129c549f0f/django/core/cache/backends/dummy.py) in Django itself is a good example of all the defaults.

I'm happy to put together a PR myself, but wanted to open this now if you have any suggestions.

ecmendenhall:
Looks like defaults were already handled in `get()` :+1:.

PR #150, opened 2015-07-16
Return empty dict from get_many()

ecmendenhall:
Fix for #149 

- Return an empty dict from django_redis.cache.RedisCache.get_many() when
  IGNORE_EXCEPTIONS is True.
- Convert omit_exception decorator to a callable. Pass it a return_value
  keyword argument to specify a default return value when
  IGNORE_EXCEPTIONS is True.


ecmendenhall:
I'm not sure why the Python 3.3/Django 1.6 build stalled out. Everything else is green: https://travis-ci.org/niwinz/django-redis/jobs/71326523

niwinz:
Already merged manually with some changes. Thanks!

PR #151, opened 2015-07-17
check for serializer class while getting/setting

smahs:
At this commit, it checks for JSONSerializer and defaults to pickle otherwise.


smahs:
Checking for wrong exception for json.loads.

PR #152, opened 2015-07-17
check for serializer class while getting/setting

smahs:
At this commit, it checks for JSONSerializer and defaults to pickle otherwise.

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -24,6 +24,8 @@ except ImportError:
 
 from redis.exceptions import ConnectionError
 from redis.exceptions import ResponseError
+from django_redis.serializers.json import JSONSerializer
+from django_redis.serializers.msgpack import MSGPackSerializer
 
 # Compatibility with redis-py 2.10.x+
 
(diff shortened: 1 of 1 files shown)


niwinz:
I'm not clearly understand the purpose of this PR, can you explain it please?

In any case thanks for contributing.

smahs:
Currently it is possible to specify a custom serializer class, but the default client does not use it. It always uses pickle to serialize objects. This patch makes set/get methods of the default client class use JSONSerializer class' dumps/loads methods if the default serializer is JSONSerializer.

niwinz:
Hmm...
The serializer is instanciated here https://github.com/niwinz/django-redis/blob/master/django_redis/client/default.py#L64 and used directly on `decode`  and `encode` methods: https://github.com/niwinz/django-redis/blob/master/django_redis/client/default.py#L312

And `decode`/`encode` are used across all library... so the serializers uses correctly the specified serializer class. Additionaly we have tests for that.

Can you provide some test case that prove the incorrect usage of the serializers by default client?

PR #153, opened 2015-07-20
Allow user to log/set logger for Ignored Exceptions

brandoshmando:
Proposed change is to allow users to log ignored exceptions so they are not failing silently. This has proven to be very useful in production. User would set the following global variables in their `settings.py` file:

`DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS` - boolean (defaults to `False`)
`DJANGO_REDIS_LOGGER` - string name/path for logger (defaults to `__name__`)

Let me know your thoughts, and if you think it looks good, I'd be more than happy to add docs.

### DIFF
--- a/django_redis/cache.py
+++ b/django_redis/cache.py
@@ -1,5 +1,6 @@
 import functools
 import warnings
+import logging
 
 from django.conf import settings
 from django.core.cache.backends.base import BaseCache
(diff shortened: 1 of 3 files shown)


brandoshmando:
I'll need to fix test cases...

niwinz:
Good work! Waitig for fixed tests for merge!

Thanks!

brandoshmando:
:+1:  Thanks!

niwinz:
Great! Tomorrow on my first time I'll review it and merge!

PR #154, opened 2015-07-22
cache.get doesn't obey ignore exceptions

diwu1989:
if the particular instance of the backend is configured to ignore exceptions but the global setting isn't, that instance should still ignore exception

### DIFF
--- a/django_redis/cache.py
+++ b/django_redis/cache.py
@@ -72,7 +72,7 @@ class RedisCache(BaseCache):
             return self.client.get(key, default=default, version=version,
                                    client=client)
         except ConnectionInterrupted:
-            if DJANGO_REDIS_IGNORE_EXCEPTIONS:
+            if self._ignore_exceptions:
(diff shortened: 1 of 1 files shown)


django_redis/cache.py
```diff
@@ -72,7 +72,7 @@ def get(self, key, default=None, version=None, client=None):
             return self.client.get(key, default=default, version=version,
                                    client=client)
         except ConnectionInterrupted:
-            if DJANGO_REDIS_IGNORE_EXCEPTIONS:
+            if self._ignore_exceptions:
```
diwu1989: the value of DJANGO_REDIS_IGNORE_EXCEPTIONS can be different from self._ignore_exceptions

diwu1989:
the travis test failure is weird, this shouldn't have different behavior on different versions of python

niwinz:
Merged! Thanks!

PR #162, opened 2015-10-16
Fix delete_many for empty generators

ostcar:
When delete_many is called with a generator instead of a list, then ```bool(keys)``` is always true, even when the generator is "empty".

But the line ```keys = [self.make_key(k, version=version) for k in keys]``` converts the generator into a list so ```bool(keys)``` does work again after the line.


niwinz:
Thanks!

PR #163, opened 2015-10-21
Update author's email

iknite:
I'm guessing that one since there is no MX records in .be DNS, 

Kudos!


niwinz:
Thanks for you time fixing it, but you are modifying the autogenerated file. Do not worry and thank for notice me. I will update it and push a new version!

iknite:
It was a minute so I'm glad it help

PR #169, opened 2015-12-22
Fix cache.clear() to respect KEY_PREFIX

### LINKED ISSUE #168: cache.clear() clears ignores KEY_PREFIX
The current implementation of clear is:
```python
    def clear(self, client=None):
        """
        Flush all cache keys.
        """
        if client is None:

        client.flushdb()
```
The last line clears the hole redis database and not only the keys starting the configured KEY_PREFIX

This happens as example, if different django caches are created with different KEY_PREFIX values.

One possible solution would be to use:
```python
    def clear(self, client=None):
        """
        Flush all cache keys.
        """
        self.delete_pattern("*")
```

ostcar:
Before this commit cache.clear() called flushdb which deletes all keys.

Fixes #168

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -283,10 +283,7 @@ class DefaultClient(object):
         """
         Flush all cache keys.
         """
-        if client is None:
-            client = self.get_client(write=True)
-
-        client.flushdb()
+        self.delete_pattern("*", client=client)
 
     def decode(self, value):
         """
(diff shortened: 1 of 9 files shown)


ostcar:
Could you give me a tip how to fix travis? I tested this PR successfully with ```python runtests.py``` like described in tests/README.txt

niwinz:
You should add the cache config to other settings  such as `test_sqlite_shard.py` among others in the same directory.

Thanks for your time!

ostcar:
To fix the tests I had to add the key-word argument ```client``` to client.sharded.ShardClient.delete_pattern```. This seems wrong, because it is not needed at this method. But I do not really understand what the ```ShardClient``` is and when client is used.

Can you give me a hint how to fix this?

niwinz:
You have added it because is part of the "protocol" but in this case is just useless. Do not worry about it.

Thanks!

PR #171, opened 2016-01-10
Include documentation and tests in source tarball.

fladi:
This allows distribution packages (e.g. Debian) to run the tests and build the
documentation from source with only the pypi source tarball.


niwinz:
Thanks!

PR #173, opened 2016-01-11
Adds the version number inside the Python module.

### LINKED ISSUE #172: Include the version number in the module
It would be great if we could access the version number directly from the Python module `django_redis`, as it is the case in most Python packages, as it is suggested in [PEP 8](https://www.python.org/dev/peps/pep-0008/#version-bookkeeping) and [PEP 396](https://www.python.org/dev/peps/pep-0396/) (even though this last one was deferred).

It could be implemented [like in Django](https://github.com/django/django/blob/master/django/__init__.py), with a `VERSION` tuple and the `__version__` string would be generated from it. Of course, other mentions of this version number (like in setup.py) should import one of these two variables.

BertrandBordage:
Fixes #172.

### DIFF
--- a/django_redis/__init__.py
+++ b/django_redis/__init__.py
@@ -10,6 +10,10 @@ else:
         return caches[alias]
 
 
+VERSION = (4, 3, 0)
+__version__ = '.'.join(map(str, VERSION))
+
+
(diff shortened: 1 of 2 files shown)


niwinz:
Thanks!

PR #188, opened 2016-04-18
Fix timeout bug.

skorokithakis:
When passing a float between 0 and 1 as the timeout (e.g. 0.5), the code would execute `int(timeout)`, which returns 0, and would never expire the key. My change sets a positive timeout to a minimum of 1.

`math.ceil(timeout)` is another course of action, but I decided not to change the current semantics too much.

If this could be changed and released as a bugfix ASAP, I would be grateful, as it is impacting production very much for us.

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -129,7 +129,10 @@ class DefaultClient(object):
         try:
             if timeout is not None:
                 if timeout > 0:
-                    timeout = int(timeout)
+                    # If "timeout" is 0.1, this would make it 0, which redis
+                    # considers as "don't expire". Instead, set it to a minimum
+                    # of 1.
+                    timeout = max(1, int(timeout))
                 elif timeout <= 0:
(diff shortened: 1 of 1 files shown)


niwinz:
Thanks!

skorokithakis:
Thanks for the quick merge! Any idea on when we can expect a release?

By the way, shoutout to [python-semantic-release](https://python-semantic-release.readthedocs.org/), it's made releasing stuff a breeze for me.

niwinz:
Thanks for the suggestion, I'll try to release it tomorrow, in any case if is important for you, and for some reason I can't do it tomorrow, ping me please ;)

skorokithakis:
Will do, thank you!

skorokithakis:
Ping!

niwinz:
@skorokithakis done! thansk!

skorokithakis:
Thank you!

PR #196, opened 2016-05-24
Add crypted serializer

onegreyonewhite:
Using cryptography lib from [this](https://github.com/pyca/cryptography).
Here in this form I use in my project. Most likely it is useful to very many people, who use django-redis in the cloud and beyond.
Perhaps you offer something to add or change - I'm ready. In my opinion this is a very useful feature.

PS: sorry for my English.

### DIFF
--- a/django_redis/serializers/crypted.py
+++ b/django_redis/serializers/crypted.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, unicode_literals
+
+from .pickle import PickleSerializer
+
+from cryptography.fernet import Fernet
+from django.conf import settings
+
+DEFAULT_KEY = r'kPEDO_pSrPh3qGJVfGAflLZXKAh4AuHU64tTlP-f_PY='
+
+
+class CryptedSerializer(PickleSerializer):
+
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


django_redis/serializers/crypted.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, unicode_literals
+
+from .pickle import PickleSerializer
+
+from cryptography.fernet import Fernet
+from django.conf import settings
```
kevin-brown: It doesn't look like this import is needed.

niwinz:
I think that this can live outside of django redis in my opinion.

Bbecause of this, django redis exposes the pluggable clients, serializers and connection management, allowing users to create more especialized plugins for extend it.

However, having that default encryption key in the source code makes the encryption completly useless, because the user can forgot configure a key and using fixed publicy known key is like no using any encryption.

Thanks for your time and your contribution, but I don't want merge this because I don't want maintain it.

django_redis/serializers/crypted.py
onegreyonewhite: Yes, there is such a thing ... We must be clear at itself in code.
In any case, it does not matter: the project author sees no sense in this commit.

PR #213, opened 2016-08-22
Added Django 1.10 to tox testing

mackeyja92:
Django 1.10 was officially released on August 1st, 2016. Adding in testing for support of Django 1.10. Also updated Django 1.9 to latest release.


niwinz:
Thanks!

PR #215, opened 2016-08-23
Document support for all supported Django versions and include 1.10 in tests

jdufresne:
You may wish to consider removing support for all except Django 1.8+ as the older versions are no longer supported or maintained upstream; including for security issues. See [supported version documentation](https://www.djangoproject.com/download/#supported-versions) for more details. If you agree I can modify the PR to remove these older versions, including any workarounds in place that would no longer be required.


niwinz:
+1 to remove support for django < 1.8 ;)

Thanks

jdufresne:
> +1 to remove support for django < 1.8 ;)

Done, thanks. All additional feedback welcome.

niwinz:
Awesome! To many compatibility stuff removed, is fantastic!

Thank you very much.

Issue #225, opened 2016-12-13
Low performance of delete_pattern


komar:
When the `delete_pattern` is being called with a pattern that will match a large number of keys then `scan_iter` is being called number of times using `SCAN`. 
The default `count` value for `SCAN` is 10 which is the number of keys being returned per call. This results in poor performance when using `delete_pattern`. `SCAN` is being called numerous times which adds lots of overhead.

pardo:
I'm facing the same performance issue with the new version using the default client

The new version seem to be calling the the delete cmd a lot more than the old one.

new version: https://github.com/niwinz/django-redis/blob/master/django_redis/client/default.py#L256

old one: https://github.com/niwinz/django-redis/blob/3.8.x/django_redis/client/default.py#L236

niwinz:
Just use the appropriate itersize number for reduce the amount of calls to scan/delete. Take care that increasing the itersize trades the number of calls to delete with the used memory, the bigger the number, the more memory will be used and less calls to delete will be done.

PR #226, opened 2016-12-13
Add support for custom count value when using delete_pattern

komar:
Solves #225


niwinz:
Many thanks! Nice work!

Only one thing before merge: can you remove, please the unrelated code to the fix? the coverage and codecov. I don't want condecov integration (I don't like much bots commenting on github issues).

niwinz:
Applied manually, thanks!

komar:
Looks like i wasn't quick enough with mods. Thank you for applying the patch manually 👍

PR #227, opened 2016-12-15
Allow to write to slave, if master is down

alanjds:
Activated only if ``OPTIONS['SLAVE_READ_ONLY']=False``.

Double-negative names are not nice, but made this way to keep the same as the Redis config option.

Implemented only on Default client and SET operation.

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -80,7 +80,7 @@ class DefaultClient(object):
 
         return random.randint(1, len(self._server) - 1)
 
-    def get_client(self, write=True, tried=[]):
+    def get_client(self, write=True, tried=[], show_index=False):
         """
         Method used for obtain a raw redis client.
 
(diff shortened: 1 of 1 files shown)


django_redis/client/default.py
```diff
@@ -61,7 +62,7 @@ def __init__(self, server, params, backend):
     def __contains__(self, key):
         return self.has_key(key)
 
-    def get_next_client_index(self, write=True):
+    def get_next_client_index(self, write=True, tried=[]):
```
niwinz: Please, don't use mutable defaults.

django_redis/client/default.py
```diff
@@ -70,25 +71,32 @@ def get_next_client_index(self, write=True):
...
+            not_tried = [i for i in range(0, len(self._server)) if i not in tried]
+            return random.choice(not_tried)
+
         if write or len(self._server) == 1:
             return 0
 
         return random.randint(1, len(self._server) - 1)
 
-    def get_client(self, write=True):
+    def get_client(self, write=True, tried=[], show_index=False):
```
niwinz: The same here, don't use mutable defaults.

alanjds: Oops. Sorry for that.

alanjds: Well... it was never mutated anyway, but better safe than sorry. Changed to tuple defaults

niwinz:
Thanks!

PR #240, opened 2017-01-29
Drop deprecated exception with typo ConnectionInterrumped

jdufresne:
Code should use ConnectionInterrupted instead. ConnectionInterrupted has been
available since version 3.4.0 released 2013-12-15. Enough time for code to have
been adjusted.


niwinz:
Completly aggree.

PR #247, opened 2017-03-18
use dill as a pickle option

suoinguon:


niwinz:
Can you explain the motivation for include dill support?

PR #252, opened 2017-04-18
use generator expression to avoid storing unnecessary list in memory

jdufresne:
https://www.python.org/dev/peps/pep-0289/


niwinz:
Thanks!

PR #253, opened 2017-04-18
Fix invalid escape sequence deprecation warning

jdufresne:
Appears during tests:

```
DeprecationWarning: invalid escape sequence \{
```

https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

> A backslash-character pair that is not a valid escape sequence now
  generates a DeprecationWarning. Although this will eventually become a
  SyntaxError, that will not be for several Python
  releases. (Contributed by Emanuel Barry in bpo-27364.)


niwinz:
Thanks!

PR #257, opened 2017-05-17
Accept generators in get_many keys

JamshedVesuna:
`get_many` iterates over `keys` twice, which will result in a StopIteration Exception. Instead, we can iterate over `keys` once, creating a map of new to old, and take advantage of the iterable nature of dictionaries when creating `args`, while preserving their order with an OrderedDict.


niwinz:
Nice! Thanks!

PR #262, opened 2017-07-27
Fix unicode cache keys on Python 2.7

linux-warrior:


encbladexp:
@niwinz is it planned to merge this? We have also this issue with Unicode Keys.

jordiecometrica:
I would like this fix too. I can work around it for now in Python 2 by making sure I encode `str` before setting a cache key.

django_redis/util.py
```diff
@@ -8,7 +8,7 @@
 
 
 @python_2_unicode_compatible
-class CacheKey(str):
+class CacheKey(object):
```
jordiecometrica: This fix seems wrong. A cache key should be a string. You should be able to call `len()`, `format`, and other stringy things on it. Doesn't this change break that?

linux-warrior: Even though `CacheKey` is no longer a string, it still has a `__str__` method (or `__unicode__` on Python 2). When we use an instance of this class as an argument for a Redis command, this method will convert it to the contained cache key. And why do you need to call `len()` or `format` on a cache key wrapper?

niwinz:
Hi @linux-warrior, please keep the PR small and focused. I don't understand the need to add a new dependency, is that mandatory for the fix?

linux-warrior:
The problem with the current code in `django_redis.client.sharded` and some other places is the usage of the `str` class, which has different meanings on Python 2 and Python 3. The usual way to resolve this and similar problems is to use a Python 2 and 3 compatibility library like `six` or `future`. Another popular option is to create a `compat.py` file in the project and define a `text_type` wrapper there, among other things. What would you recommend doing?

niwinz:
I prefer keep the things simple and not adding an additional dependency. If this is just one thing, just define a compat.py module and there the need code.

Thanks

linux-warrior:
Done.

jordiecometrica:
@niwinz Sorry to be a bother, but will you be doing a new release of django-redis soon? If not, I can just point my requirements.txt to the git repo instead of pypi.

PR #267, opened 2017-09-22
Avoid error when timeout is less than 1ms

rrafal:
Redis requires that PX option of SET command to be a positive integer. If PX is 0, then it returns "invalid expire time in set" error.

In my application, I call `django_redis.client.DefaultClient.set` method. Timeout is set to a small random value. When timeout happens to be less than 0.001, `DefaultClient.set` first checks that timeout is positive, then converts it to 0 milliseconds. Obviously, PX gets set to 0 and I get back an exception.

It would be better if `DefaultClient.set` converted timeout to milliseconds first, then check if it is positive.


niwinz:
Thanks

PR #274, opened 2017-10-20
Fix tests and tox configuration to run successfully

jdufresne:
Previously, tox was executing 0 tests.

Reorganize the envlist in terms of Django versions. Much easier to keep organized and (when it is time) drop outdated version. The tox and Travis matrix now match.

Required adding missing dependency to tox configuration.

Add missing supported Python version to trove classifiers. Python 3.3 is still being tested by tox and Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,12 +35,12 @@ before_script:
   - flake8
 
 script:
-  - (cd tests && python runtests.py)
-  - (cd tests && python runtests-sharded.py)
-  - (cd tests && python runtests-herd.py)
-  - (cd tests && python runtests-json.py)
-  - (cd tests && python runtests-msgpack.py)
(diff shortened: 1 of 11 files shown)


niwinz:
The main problem with tox is that i'm not using it, so fixing it only corresponds to the community :( i will merge it right now, but i think i will consider remove that from the repository, just because i'm not using it.

jdufresne:
I'm fine either way. I simply noticed the tox.ini wasn't working so fixed it.

Just curious, when making changes, what do you use to verify all tests in the matrix pass? Or do you only rely on Travis for that?

niwinz:
I'm only pass them locally for one python version, and then just relly on travis for the rest of the matrix. If some environment is broken, i just try to mount that concrete environment and fix it. But in my past experience this is never needed.

PR #275, opened 2017-10-20
Introduce isort to manage and organize imports in files

jdufresne:
Use Django's configuration:

https://github.com/django/django/blob/0f722d865ee556816c36ae6c6bf229904b5142dd/setup.cfg#L10-L17

Ensures all files use a consistent import style. Removes the need for contributors to think or manage the imports manually. Added Travis CI check to make sure imports remains the same over time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ install:
 
 before_script:
   - flake8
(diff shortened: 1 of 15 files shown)


niwinz:
Merged manually

PR #276, opened 2017-10-21
Use Django's override_settings to override cache settings in tests

jdufresne:
Avoids half re-implementing it some other way. Ensures the Django `setting_changed` signal fires.


niwinz:
Merged manually.

PR #277, opened 2017-10-21
Escape prefix and version in glob expressions

jdufresne:
A prefix containing a * should be escaped and not interpreted as a glob pattern.


niwinz:
Merged manually, many thanks for all the work

PR #279, opened 2017-10-24
compatibility with migration from py2 to py3 while preserving cache

thinkt4nk:
When I migrate from py2 to py3, py3's `pickle.loads` fails when it attempts to decode with ascii encoding as [its documentation](https://docs.python.org/3/library/pickle.html#pickle.loads) details. 

I'm not entirely sure why the existing codebase works on py3 with py3 serialized objects and not py2. It may be that py3 somehow serializes with 7-bit encoding, so that it's ascii compatible, and py2 with 8-bit bytes encoding? There's [an issue](https://github.com/niwinz/django-redis/issues/184) that was opened for this problem, and while the reporter was able to resolve by flushing the cache, we need to preserve our cache through our migration from py2 to py3, so I wanted to fix this issue.

### DIFF
--- a/django_redis/serializers/pickle.py
+++ b/django_redis/serializers/pickle.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import absolute_import, unicode_literals
+from ..compat import PY3
 
 # Import the fastest implementation of
 # pickle package. This should be removed
@@ -33,4 +34,6 @@ class PickleSerializer(BaseSerializer):
         return pickle.dumps(value, self._pickle_version)
 
     def loads(self, value):
+        if PY3:
+            return pickle.loads(force_bytes(value), fix_imports=True, encoding='bytes')
         return pickle.loads(force_bytes(value))


niwinz:
python3 uses variable length encoding strings, if your strings contains only ascii compaible chars, it will encode it as utf8 in the same way as it will be in ascii (utf8 is compatible with ascii for the ascii subset of chars).

Because of that, when you on python2 serializes a byte string (normal string in python2) with non ascii chars, it will fail on load on python3, but if you correctly uses unicode strings for that cases, it will work as expected.

Considering so, i think this fix is just a workaround for your specific case and I think that applying this change for all the django-redis uses seems like a not soo good approach.

I suggest create a script that just iterates over all keys in the database , load the content with the workaround and then save it again, all using python3, ensuring that resulting encoding will load on your python3 apps.

thinkt4nk:
That's close to what I figured was happening, but I think I'm still unclear on just *why* this fix is not a good approach. I think that pushing downstream to an undocumented one-off migration process is a worse approach. It's obviously bitten others before, and I'm sure it'll bite others to come. Anyway, we've forked for our own purposes, so I don't care too much what you choose to do. Thanks for looking!

niwinz:
This is a not a good general porpuse solution because it solves some issues but may cause other issues. 

A great example that the keys of the dits, if in python2 they are strings (byte strings) with ascii compatible content, with the default aproach in python3 they will be strings (unicode strings) but with your proposal they will be as bytes and that also can break some other code and will imply a manual intervention.

thinkt4nk:
default approach in python3 they will be strings when you run it through `_force_bytes`?

niwinz:
I'm talking about the result of pickle.loads, not the input (that is passed through the force_bytes)

niwinz:
This is an example:
```
$ python2
Python 2.7.14 (default, Sep 20 2017, 01:25:59) 
[GCC 7.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pickle, io
>>> pickle.dump({"foo": "bar"}, io.open("/tmp/testfile", "wb"))
```

```
$ python3
Python 3.6.2 (default, Jul 20 2017, 03:52:27) 
[GCC 7.1.1 20170630] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pickle, io
>>> pickle.load(io.open("/tmp/testfile", "rb"), encoding="bytes")
{b'foo': b'bar'}
>>> pickle.load(io.open("/tmp/testfile", "rb"))
{'foo': 'bar'}
```

jarussi:
For anyone that is having a similar issue, instead of forking django-redis, I created a custom serializer and passed as an argument in the cache options settings. What did the trick for me was a little bit different. Here is my serializer:
```
import six
from django_redis.serializers.pickle import PickleSerializer

try:
    import cPickle as pickle
except ImportError:
    import pickle


class MyPickleSerializer(PickleSerializer):

    def loads(self, value):
        if six.PY3:
            return self._loads_py3(value)
        return super(MyPickleSerializer, self).loads(value)

    def _loads_py3(self, value):
        return pickle.loads(
            value,
            fix_imports=True,
            encoding='latin1'
        )
```
Here is my settings for the cache. The Pickle Version is also important.
```
CACHES = {
    'default': {
        'BACKEND': 'django_redis.cache.RedisCache',
        'LOCATION': 'localhost:6379',
        'OPTIONS': {
            'CLIENT_CLASS': 'django_redis.client.DefaultClient',
            'SERIALIZER': (
                'path.to.my.serializer.MyPickleSerializer'
            ),
        },
        'PICKLE_VERSION': 2,
}
```

PR #278, opened 2017-10-21
Remove broken fakeredis support and testing

jdufresne:
fakeredis is a very incomplete implementation of the Python Redis interface. Many django-redis tests fail or are skipped when running tests against fakeredis. The situation doesn't appear to be improving. The brokenness causes non-zero maintenance. To reduce maintenance, drop support for the broken features.

Fixes #270

PR #280, opened 2017-10-27
Fix merge cleanups causing Travis test failures

jdufresne:
* Remove missed references to fakeredis. Remained after merge of #278.
* Fixed missed import ordering. Remained after merge of #275.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,8 @@ install:
   - pip install -r requirements.txt
 
 before_script:
-  - (cd django_redis; flake8)
-  - (cd django_redis; isort --check-only --diff)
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -32,8 +32,8 @@ install:
   - pip install -r requirements.txt
 
 before_script:
-  - (cd django_redis; flake8)
-  - (cd django_redis; isort --check-only --diff)
+  - flake8 django_redis tests
+  - isort --check-only --diff --recursive django_redis tests
```
niwinz: Oh thanks, i took a look on the --help but didn't see that option. That was a workaround :P

PR #281, opened 2017-11-09
add error threshold

markddavidoff:
Please consider this new feature. This adds error threshold with a configurable time window and cooldown. If this looks good, I'll update the docs as well.

### DIFF
--- a/django_redis/cache.py
+++ b/django_redis/cache.py
@@ -1,17 +1,21 @@
 import functools
 import logging
+import time
+from threading import RLock
 
 from django.conf import settings
 from django.core.cache.backends.base import BaseCache
 
-from .util import load_class
 from .exceptions import ConnectionInterrupted
-
+from .util import load_class
 
 DJANGO_REDIS_IGNORE_EXCEPTIONS = getattr(settings, "DJANGO_REDIS_IGNORE_EXCEPTIONS", False)
 DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = getattr(settings, "DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS", False)
 DJANGO_REDIS_LOGGER = getattr(settings, "DJANGO_REDIS_LOGGER", False)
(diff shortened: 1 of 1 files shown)


markddavidoff:
@niwinz When you have time, input would be appreciated :).

niwinz:
Please can you explain this PR a little bit? maybe some documentation also will be awesome.
At simple view i don't understand the purpose of it, so this is not obvious at quick view.

markddavidoff:
This pr adds the ability to skip cache calls to redis for x seconds when there are y errors within z seconds.

the x, y and z are controlled by the 3 configs (respectively): DJANGO_REDIS_EXCEPTION_THRESHOLD, DJANGO_REDIS_EXCEPTION_THRESHOLD_TIME_WINDOW, DJANGO_REDIS_EXCEPTION_THRESHOLD_COOLDOWN

niwinz:
Hmm nice work then, but i think we need some documentation for that part, maybe some little section on "Advanced usage", with an example like you have said in previous comment. And some tests are also mandatory

markddavidoff:
@niwinz Definitely, just wanted to gauge if you thought this was a good idea first.

niwinz:
I think that this can be done out of this project, django-redis allows create user defined clients with user defined behavior overwriting the default behavior. This work is very interesting but adds a lot of complexity that, sincerelly i dont want on the main client.

markddavidoff:
@niwinz Sorry for taking so long on this, I had completely forgot about it.

I can see your point, however I have a reason to justify this complexity that made me think it would be useful as part of this lib (perhaps I should have started with that but It had slipped my mind).

When you have multiple cache calls in a request, both timeouts and ignore exceptions are set and redis becomes unavailable, the request hangs until all the timeouts resolve. Through multiple requests, each with multiple cache hits (such as someone spamming refresh) this can ended up hanging the whole server.

I want ignore_exceptions to allow the cache to function as if there was no data returned on error, but still allow for occasional network stutters by having a timeout set to nonzero.

The thresholding I suggest allows you to fail fast when you have a redis outtage and prevent hanging.

PR #288, opened 2017-11-28
Be more explicit about types in the JSONSerializer

jdufresne:
Always return utf-8 encoded bytes in dumps. The Redis connection will pass it as is. Always expected utf-8 encoded bytes in loads. This makes the functions inverses of one another.

Using `force_bytes()`/`force_text()` implies an unknown input/output, but the input/output is always known. It should always be bytes.


jdufresne:
Rebased to resolve merge conflicts.

PR #302, opened 2017-12-27
Remove test suite's database dependency

jdufresne:
The test suite never actually hits the database, should therefore not configure one. As django-redis is cache only, the database is not involved in testing. Use Python `unitest.TestCase` instead of Django's
`TestCase`. It is slightly faster as it doesn't begin a database transaction.

- https://docs.djangoproject.com/en/2.0/topics/testing/tools/#provided-test-case-classes
- https://docs.djangoproject.com/en/2.0/topics/testing/tools/#simpletestcase


niwinz:
Many thanks for all you work!

PR #305, opened 2018-01-08
atomic incr() if the key is not set (supporting only DefaultClient)

maheshgawali:
**based on discussion here**
https://github.com/niwinz/django-redis/issues/4#issuecomment-314082634

**excerpt below**
I have a usecase where incr is intended to be done by mulitple processes and using set (to init keys) poses a risk of multiple processes NOT finding the key and then setting it to a default value and then incrementing.
```python
value = redis_conn.get(redis_key)
if value:
    redis_conn.incr(redis_key, delta=count)
else:
    redis_conn.set(redis_key, delta=count, timeout=PREDECIDED_TIMEOUT)
```
what I want is:
```python
redis_conn.incr(redis_key, delta=count, ignore_key_check=True)
redis_conn.expire(redis_key, timeout=PREDECIDED_TIMEOUT)
```
Note that none of the existing calls to `incr` will have to be changed, only the explicit calls which use `ignore_key_check` will go through the new flow.

I have been using this for a while and it works for me, however, please let me know your review comments and any improvements if required.


maheshgawali:
All tests are passing for me on local, is CI failing because of PR across forks ?

tarzzz:
Hi @maheshgawali , Are there any plans to get this PR merged some time soon. I am also hitting a similar issue and would be great to get an update on this.. !!

scjody:
We should tag @niwinz in case he hasn't seen this yet.

niwinz:
oh, I had not seen it sorry, thanks for tagging.
It is ok for this PR, so i will merge and release new version this monday probably. If not, just pingme, i have a lot of stuff to do, so i just can forget

tarzzz:
@niwinz reminder for this.. !!

PR #313, opened 2018-03-05
adding support for INCRBYFLOAT (DefaultClient)

maheshgawali:
As per the discussion [here](https://github.com/niwinz/django-redis/issues/311) adding support for `incr_by_float` (supporting only `DefaultClient`)
Redis Documentation link [here](https://redis.io/commands/incrbyfloat).
@niwinz : If `incr` should be handling `incr_by_float` using extra parameter, please leave a comment and I'll update this PR accordingly.

Thanks @niwinz for an awesome lib


niwinz:
Thanks for your work, but as i said in the related issue, we should not add new methods to the client. If you implement it on the `incr` method, i will merge it.

PR #318, opened 2018-03-25
Update docs to reflect supported versions

### LINKED ISSUE #316: django-redis==3.8.4 fails with django 1.11
django-redis==3.8.4 is using `get_cache` [1] which was removed in django 1.9 [2]

This throws the following exception:

```
InvalidCacheBackendError: Could not find backend 'django_redis.cache.RedisCache': cannot import name get_cache
```

[1] https://github.com/niwinz/django-redis/blob/3.8.x/django_redis/__init__.py#L3
[2] https://docs.djangoproject.com/en/2.0/releases/1.9/#features-removed-in-1-9

kevin-brown:
* Remove version 3.8.x from documentation

  This version is clearly no longer supported, as it hasn't been
touched since 2015 and the documentation no longer lines up with
what that version actually supports. At this point, it makes more
sense to remove any references to that version than it does to
keep the documentation up to date with a version from 3 years ago.

* Bump docs stable version to 4.9.0

  This version was released last month and while the live docs appear
to reflect this, the source docs do not reflect the new version.

This should close #316 and close #317 since they would officially be unsupported.

### DIFF
--- a/doc/content.adoc
+++ b/doc/content.adoc
@@ -39,17 +39,16 @@ Supported django-redis versions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 - Supported stable version: *4.8.0*
-- Supported stable version: *3.8.4*
 
 
 How version number is handled
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Versions like _3.6_, _3.7_, ... are considered major releases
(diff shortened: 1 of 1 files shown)


niwinz:
Ty

kevin-brown:
@niwinz if you can re-deploy the documentation, it'll be up to date with the source and we won't be seeing  tickets in the future about the versions not being supported.

PR #321, opened 2018-04-16
Add dill serializer

svenauhagen:
Hi,

this is a dill serializer.
I do need it because I use django model utils to track fields.
That plugin does not work with pickle but dill works on serializing models with it:

https://github.com/jazzband/django-model-utils/issues/83

Best and thanks
Sven

### DIFF
--- a/django_redis/serializers/dill.py
+++ b/django_redis/serializers/dill.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
-import dill
-
 from __future__ import absolute_import, unicode_literals
+
+import dill
 
 from django.core.exceptions import ImproperlyConfigured
 
 from .base import BaseSerializer
+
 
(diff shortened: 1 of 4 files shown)


eyesee1:
dill needs to be added to requirements.txt

jdufresne:
I think it would be good to add tests similar to the other serializers.

niwinz:
Tests are mandatory for new features.

svenauhagen:
Hi I added some test for dill.
Let me know if that is the correct way of adding the tests for the project.

peterbe:
If someone knows they want `django-redis` with `dill` they'd install it with `pip install "django-redis[dill]"`. To make that work you'd just add to `setup.py`...:

```python
  extras_require={
      'dill': ['dill'],
  },
```

peterbe:
Can this PR also come with a change to the docs?

PR #326, opened 2018-07-19
Fix get method exception catching

TheOldMop:
Example:
```
DJANGO_REDIS_IGNORE_EXCEPTIONS = True
CACHES = {
    "default": {
        "OPTIONS": {
            "IGNORE_EXCEPTIONS": False,
        }
    },
}
```
Stop redis and check:
`caches['default'].set('foo', 'bar')`
`ConnectionError: Error 61 connecting to localhost:32768. Connection refused`.
`caches['default'].get('foo')`
Exception ignored

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -659,6 +659,43 @@ class DjangoOmitExceptionsTests(unittest.TestCase):
         self.assertEqual(self.cache.get("key", default="default"), "default")
 
 
+class DjangoOmitExceptionsPriority1Tests(unittest.TestCase):
+    def setUp(self):
+        self._orig_setting = django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS
+        django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS = False
+        caches_setting = copy.deepcopy(settings.CACHES)
+        caches_setting["doesnotexist"]["IGNORE_EXCEPTIONS"] = True
+        cm = override_settings(CACHES=caches_setting)
+        cm.enable()
+        self.addCleanup(cm.disable)
+        self.cache = caches["doesnotexist"]
+
+    def tearDown(self):
+        django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS = self._orig_setting
+
+    def test_get(self):
+        self.assertIsNone(self.cache.get("key"))
+
+
+class DjangoOmitExceptionsPriority2Tests(unittest.TestCase):
+    def setUp(self):
+        self._orig_setting = django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS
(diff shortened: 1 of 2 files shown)


jdufresne:
Thanks! Could you add a test that demonstrates the fix?

TheOldMop:
Okey! I will try do this in the next  few days.

niwinz:
Thanks

tests/test_backend.py
```diff
@@ -659,6 +659,43 @@ def test_get(self):
         self.assertEqual(self.cache.get("key", default="default"), "default")
 
 
+class DjangoOmitExceptionsPriority1Tests(unittest.TestCase):
+    def setUp(self):
+        self._orig_setting = django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS
+        django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS = False
+        caches_setting = copy.deepcopy(settings.CACHES)
+        caches_setting["doesnotexist"]["IGNORE_EXCEPTIONS"] = True
```
jdufresne: I think this was intended to be:

```py
caches_setting["doesnotexist"]["OPTIONS"]["IGNORE_EXCEPTIONS"] = True
```

("OPTIONS" key is missing)

tests/test_backend.py
```diff
@@ -659,6 +659,43 @@ def test_get(self):
...
+        cm = override_settings(CACHES=caches_setting)
+        cm.enable()
+        self.addCleanup(cm.disable)
+        self.cache = caches["doesnotexist"]
+
+    def tearDown(self):
+        django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS = self._orig_setting
+
+    def test_get(self):
+        with self.assertRaises(KeyError):
```
jdufresne: This test fails for me on master as a `KeyError` is never raised.

Further, the cache should *never* raise a `KeyError`. From the [Django docs](https://docs.djangoproject.com/en/2.1/topics/cache/#django.core.caches.cache.get). So, for this test to be valid, I think this should be expecting some other kind of exception.

tests/test_backend.py
```diff
@@ -659,6 +659,43 @@ def test_get(self):
         self.assertEqual(self.cache.get("key", default="default"), "default")
 
 
+class DjangoOmitExceptionsPriority1Tests(unittest.TestCase):
+    def setUp(self):
+        self._orig_setting = django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS
+        django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS = False
```
jdufresne: As `DJANGO_REDIS_IGNORE_EXCEPTIONS` is a setting, can this use the same `override_settings()` context manager instead of manipulating the module directly?

TheOldMop: I fix test's in this PR https://github.com/niwinz/django-redis/pull/335/files

PR #329, opened 2018-08-29
Add note about ConnectionPool kwargs

oychang:
Adds an example of how to pass `kwargs` to a Connection and reword the `ConnectionPool` example to make it clear that as the `ConnectionPool` example states, "Any additional keyword arguments are passed to the constructor of connection_class."

This behavior is because of the `dict.update()` [here](https://github.com/niwinz/django-redis/blob/6d7bfec1e29480593eb9afe84b4c94ae5586c30f/django_redis/pool.py#L107), and it looks like the behavior has always been this way. However, a rewording in #74 made it seem like the only valid key was `max_connections`.


niwinz:
❤️

PR #335, opened 2018-10-02
fix omit exceptions priority tests

TheOldMop:


TheOldMop:
Don't know why but in travis test fail 8( If run tests local it ok. Can may be some one know why?

TheOldMop:
Finally fix problem was in tests config

tests/test_sqlite_herd.py
```diff
@@ -10,11 +10,11 @@
             'CLIENT_CLASS': 'django_redis.client.HerdClient',
         }
     },
-    'doesnotexist': {
-        'BACKEND': 'django_redis.cache.RedisCache',
-        'LOCATION': '127.0.0.1:56379:1',
-        'OPTIONS': {
-            'CLIENT_CLASS': 'django_redis.client.HerdClient',
+    "doesnotexist": {
+        "BACKEND": "django_redis.cache.RedisCache",
```
jdufresne: Why are these files changing? It isn't clear why that is necessary for the tests to pass.

tests/test_backend.py
```diff
@@ -674,6 +675,7 @@ def tearDown(self):
         django_redis.cache.DJANGO_REDIS_IGNORE_EXCEPTIONS = self._orig_setting
 
     def test_get(self):
+        assert self.cache._ignore_exceptions is True
```
jdufresne: The test suite is using unittest style asserts, not the assert keyword. What do think about using `self.assertIs(self.cache._ignore_exceptions, True)` to remain consistent.

Does the test `DjangoOmitExceptionsTests` need to be updated too? It is also missing the `OPTIONS` key and doesn't have this assert.

tests/test_sqlite_herd.py
jdufresne: Oh I see. It was changed to use a bogus port. Is that right? :+1:

tests/test_sqlite_herd.py
```diff
@@ -10,11 +10,11 @@
...
+        "LOCATION": "redis://127.0.0.1:56379?db=1",
+        "OPTIONS": {
+            "CLIENT_CLASS": "django_redis.client.DefaultClient",
```
jdufresne: This should probably remain `HerdClient` though as this is specific for herd tests.

niwinz:
Already merged, thanks.

PR #336, opened 2018-10-02
add touch method

TheOldMop:


jdufresne:
Could this be done using the native `TOUCH` command? It would avoid all race conditions between `.get()` and `.set()`.

https://redis.io/commands/touch

TheOldMop:
No.
For django `touch` set new expiration time for key:
`cache.touch('a', 10)`
`True`
For redis `touch`  alters the last access time of a key(s)

They have different meaning.

jdufresne:
Thanks.  Maybe we could use `EXPIRE` or `EXPIREAT` then. They don't have a race condition and can work on existent or non-existent keys.

```
127.0.0.1:6379> expire mykey 10
(integer) 0
127.0.0.1:6379> set mykey myvalue
OK
127.0.0.1:6379> expire mykey 10
(integer) 1
```

TheOldMop:
I missed this method ( Can fix it tomorrow

django_redis/cache.py
```diff
@@ -150,3 +150,14 @@ def lock(self, *args, **kwargs):
     @omit_exception
     def close(self, **kwargs):
         self.client.close(**kwargs)
+
+    @omit_exception
+    def touch(self, key, timeout=None, version=None):
+        try:
+            return self.client.touch(key, timeout=timeout, version=version)
+        except ConnectionInterrupted as e:
```
jdufresne: Isn't this exception already handled the `omit_exception`? Why does it need to be repeated?

django_redis/client/default.py
```diff
@@ -540,3 +540,19 @@ def close(self, **kwargs):
...
+        """
+        Sets a new expiration for a key.
+        """
+
+        if client is None:
+            client = self.get_client(write=True)
+
+        key = self.make_key(key, version=version)
+
+        if client.exists(key):
```
jdufresne: Can we drop this check? My understanding is `expire` will ignore keys that doesn't exist.

Could use a test for the key-doesn't-exist case.

tests/test_backend.py
```diff
@@ -635,6 +635,30 @@ def test_master_slave_switching(self):
         except NotImplementedError:
             pass
 
+    def test_touch_zero_timeout(self):
+        self.cache.set("test_key", 222, timeout=10)
+
+        self.cache.touch("test_key", 0)
```
jdufresne: I think it would be useful to assert the return value of `touch`:

```py
self.assertIs(res, True)
```

niwinz:
Thanks!

niwinz:
Seems like some tests fails on master related to touch method, can you check this please?

TheOldMop:
I can check but only around 20:00 UTC +3

PR #341, opened 2018-11-08
Fix add() returns None instead of False

krisrp:


michael-k:
ref niwinz/django-redis#340

tests/test_backend.py
```diff
@@ -297,9 +297,10 @@ def test_timeout_tiny(self):
 
     def test_set_add(self):
         self.cache.set("add_key", "Initial value")
-        self.cache.add("add_key", "New value")
-        res = cache.get("add_key")
+        res = self.cache.add("add_key", "New value")
+        self.assertFalse(res)
```
jdufresne: `assertIs(res, False)` would be more precise as `assertFalse(None)` passes.

jdufresne:
I think it would be useful to add a test for the `True` case as well. I don't think such a test exists yet.

django_redis/client/default.py
```diff
@@ -185,6 +185,8 @@ def add(self, key, value, timeout=DEFAULT_TIMEOUT, version=None, client=None):
 
         Returns ``True`` if the object was added, ``False`` if not.
         """
+        if self.has_key(key):
+            return False
         return self.set(key, value, timeout, version=version, client=client, nx=True)
```
jdufresne: There is likely a race condition between the `has_key` & `set` call. Maybe this should use the Redis `SETNX` command instead.

https://redis.io/commands/setnx

niwinz: Agree, the current proposed approach can cause race conditions.

krisrp:
Added test for the `True` case.

django_redis/client/default.py
niwinz: yeah, seems like the nx should be default option

jdufresne: Why do we need the `.has_key()` at all then? Shouldn't `.set(..., nx=True)` be enough? Why add additional round trips to Redis?

tests/test_backend.py
```diff
@@ -298,11 +298,15 @@ def test_timeout_tiny(self):
 
     def test_set_add(self):
         self.cache.set("add_key", "Initial value")
-        self.cache.add("add_key", "New value")
-        res = cache.get("add_key")
+        res = self.cache.add("add_key", "New value")
+        self.assertEqual(res, False)
```
jdufresne: I would use `assertIs(res, False)` to ensure res is actually false and not just falsey. Same for `assertIs(res, True)`.

From the [Python docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue):

>  assertTrue(expr, msg=None)
> assertFalse(expr, msg=None)
>
> Note that this is equivalent to `bool(expr) is True` and not to `expr is True` (use `assertIs(expr, True)` for the latter).

krisrp: Whoops. I must have misread your initial comment. Thanks for your patience and kind clarification.

django_redis/client/default.py
krisrp: `.set(..., nx=True)` isn't returning False like it should. Unfortunately I don't have enough time to look more into this until the weekend. It looked like a quick fix initially, but evidently it's not.

My break time is up, but I looked briefly into it: `redis-py`'s `setnx` returns False when it fails, and when the `nx` flag is set it calls

https://github.com/niwinz/django-redis/blob/f411c553a7749013e4879ebf08c93daf222346f2/django_redis/client/default.py#L137

So it's probably something here that's not behaving as it should. The code of `has_key` looks fine to me though.

jdufresne:
This work has been continued in #406 . Thanks for the original patch!

PR #343, opened 2018-11-15
Pin redis version to 2.10.6

drgarcia1986:
To avoid some trouble with Redis [3.x series](https://github.com/andymccurdy/redis-py#upgrading-from-redis-py-2x-to-30)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     install_requires=[
         "Django>=1.11",
-        "redis>=2.10.0",
+        "redis>=2.10.0,<3.0.0",
     ],
     zip_safe=False,
     include_package_data=True,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -22,7 +22,7 @@
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     install_requires=[
         "Django>=1.11",
-        "redis>=2.10.0",
+        "redis==2.10.6",
```
jdufresne: I think a range like `<3.0.0` would be more accurate and usable.

drgarcia1986: LGTM 👍 
Done ✅

michaeljonaitis:
Yas please

jdufresne:
An alternative is PR #344 which adds support for newer redis-py and is backwards compatible with older versions.

Maks3w:
Or both. Do a minor release for this and feature release for the compatibility PR

niwinz:
Released a intermediary version with this change (4.9.1)
Thanks.

PR #344, opened 2018-11-16
Add support redis-py 3.0.0

### LINKED ISSUE #342: django-redis is incompatible with redis-py >= 3
From the redis-py release notes:

> Only bytes, strings and numbers (ints, longs and floats) are acceptable for keys and values.

See https://github.com/andymccurdy/redis-py/blob/9b03af26dc829beea232a3248768de933f4c3b67/CHANGES#L9-L15

The `CacheKey` class used by django-redis is no longer supported.

I'm happy to open a PR, but I wanted to ask first if it's better to drop `CacheKey` or to cast keys to strings right before sending them to redis-py?  The latter would be more work, but retain the intended use of `CacheKey`:

> A stub string class that we can use to check if a key was created already.

jdufresne:
- Make CacheKey inherit from `str` to fix exception "Only bytes, strings and numbers (ints, longs and floats) are acceptable for keys and values."
- `.exists()` now returns and integer, so update `.has_key()`.
- Remove erroneous tests so all tests pass again.
- Fix flake8 errors.
    
Fixes #342

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,37 +5,71 @@ cache: pip
 matrix:
   include:
     - python: 2.7
-      env: TOXENV=py27-django111
+      env: TOXENV=py27-django111-redis2
+    - python: 2.7
+      env: TOXENV=py27-django111-redislatest
     - python: 3.4
-      env: TOXENV=py34-django111
-    - python: 3.5
-      env: TOXENV=py35-django111
-    - python: 3.6
-      env: TOXENV=py36-django111
+      env: TOXENV=py34-django111-redis2
     - python: 3.4
-      env: TOXENV=py34-django20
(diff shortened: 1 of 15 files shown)


michael-k:
Extending `tox.ini` and `.travis.yml` to run tests against redis < 3 as well as >= 3 would be nice.

jdufresne:
@michael-k Good call! In the latest version I've updated tox & Travis to test against both redis-py 2.x and 3.x. Thanks for the suggestion.

jdufresne:
@TheOldMop I've incorporated your test fixes from #335 so we can verify all tests pass. Thanks!

jdufresne:
Thanks. Looks mostly identical.

niwinz:
Nice work! Thank you all.

PR #345, opened 2018-11-16
Clean up test skipping

jdufresne:
Remove unnecessary skip.

Use preferred self.skipTest() instead of returning or unittest.SkipTest.


jdufresne:
I have rebased so all tests are now green.

PR #348, opened 2018-11-16
Replace load_class with Django's import_string

jdufresne:
The local function `load_class()` duplicates Django's `import_string()`. Simply reuse it instead of re-implementing it.

https://docs.djangoproject.com/en/2.1/ref/utils/#module-django.utils.module_loading


jdufresne:
Rebased. The test failures are due to the new touch issue.

niwinz:
Yeah, i know. Thanks!

PR #350, opened 2018-11-19
fix touch method for sharded

TheOldMop:


TheOldMop:
Hm seems another problem (

TheOldMop:
I will try to increase timeout in this test, may be that solve problem

TheOldMop:
https://github.com/niwinz/django-redis/pull/351

PR #355, opened 2018-11-25
fix sleep duration

TheOldMop:


jdufresne:
Can you explain why `time.sleep(2)` is insufficient for `self.cache.touch("test_key", 2)` to timeout? It is not clear to me.

TheOldMop:
I think problem with time synchronization between app and redis. Test fall some times on different python/redis version. All other tests i see in project use +1 second for sleep before check

PR #381, opened 2019-04-25
Add more context to warnings emitted from Django when unpickling objects

Tenzer:
When upgrading Django, Django will emit a warning whenever an object is being
unpickled which was created with a different version of Django. This warning
isn't very easy to action as it doesn't contain any context about what object
it is or where it may have got the object from. The warning might look
something like this:

> WARNING [2019-04-25 09:41:48,525] warnings: /path/to/virtualenv/lib/python3.7/site-packages/django_redis/serializers/pickle.py:35: RuntimeWarning: Pickled model instance's Django version 2.1.8 does not match the current version 2.2.
>   return pickle.loads(value)

This change will loop through each warning emitted during the deserialization
of the cached item and append the cache key to the warning which at least would
give some context as to why that is happening, so the warning message instead
becomes:

> WARNING [2019-04-25 09:41:48,525] warnings: /path/to/virtualenv/lib/python3.7/site-packages/django_redis/serializers/pickle.py:35: RuntimeWarning: Pickled model instance's Django version 2.1.8 does not match the current version 2.2. (cache key: :1:user)
>   return pickle.loads(value)

given the cache key `user`.

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, unicode_literals
 import random
 import re
 import socket
+import warnings
 from collections import OrderedDict
 
 from django.conf import settings
@@ -208,7 +209,7 @@ class DefaultClient(object):
         if value is None:
             return default
 
-        return self.decode(value)
+        return self.decode(value, key)
 
     def persist(self, key, version=None, client=None):
         if client is None:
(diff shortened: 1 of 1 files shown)


terencehonles:
This will happen anywhere you use `Pickle` and only if you are pickling Django model instances. I don't believe this belongs in the `django-redis` since this will have a runtime effect and is probably not generally a problem. If you need to track this down the stack will show where's coming from in the code, and then you can apply the warnings context manager you provided in this PR to find out what key(s) might be a problem if the keys are dynamically constructed in the code.

PR #383, opened 2019-05-11
Bulk delete in DefaultClient delete_pattern

Yiling-J:
Delete keys one by one could be slow, so bulk delete them.

Split the scan_iter into chunks, and bulk delete each chunk.


django_redis/util.py
```diff
@@ -15,3 +15,27 @@ def original_key(self):
...
+        result = {next(iterator, stop) for i in range(size)}
+        if stop in result:
+            result.remove(stop)
+            yield result
+            break
+        yield result
+
+
+def delete_pattern(params, client):
+    chunks = chunk_iter(client.scan_iter(**params), 500, None)
```
SerhiyRomanov: Hello @Yiling-J ! Thank you for this pull request! 
One thing, i noticed in this line - you pass size=500 for function chunk_iter(), but count=10 will be passed (via **params) to client.scan_iter() function. Because default values for DJANGO_REDIS_SCAN_ITERSIZE is 10.
So, it leads to having 50 queries to Redis for getting needed 500 keys.
Please, see screenshot for better understanding. 
image

I guess, would be better to do this in suggested way and also increase DJANGO_REDIS_SCAN_ITERSIZE up to 500 or even 1000. As i see, variable DJANGO_REDIS_SCAN_ITERSIZE only are used with delete_pattern() function

```suggestion
    chunks = chunk_iter(client.scan_iter(**params), params.get('count', DJANGO_REDIS_SCAN_ITERSIZE), None)
```

Yiling-J: Hi Serhiy:
Actually `500` is the delete chunk size, so we can bulk delete, not scan chunk size. That means, if we have 1000 keys to delete, we only need to call `client.delete` twice. But for scan `scan_iter`, still using DJANGO_REDIS_SCAN_ITERSIZE, default 10.

`500` is fix now, We can also turn that into a setting, if needed.

Yiling-J: also clear in you screenshot, `scan` use 10 , and `delete` after `scan` use 500

SerhiyRomanov: Thank you for explaining, i totally understand how it works. I simply want to point out that if the different count parameter are used for both getting list of keys and chunk_iter() function, we will have redundant queries like on screenshot.
This is why I proposed changes

PR #385, opened 2019-06-21
Update setup to include license at top-level.

briandailey:
This makes it work with automated tools like `pip-licenses` that scan requirements.


jdufresne:
Thanks!

PR #393, opened 2019-09-25
Forward thread_local param to redis client

sindrig:
I need to forward an acquired lock to a worker thread that should release it. From redis-py docs:

        In some use cases it's necessary to disable thread local storage. For
        example, if you have code where one thread acquires a lock and passes
        that lock instance to a worker thread to release later. If thread
        local storage isn't disabled in this case, the worker thread won't see
        the token set by the thread that acquired the lock. Our assumption
        is that these cases aren't common and as such default to using
        thread local storage.

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -252,13 +252,18 @@ class DefaultClient:
         sleep=0.1,
         blocking_timeout=None,
         client=None,
+        thread_local=True,
     ):
         if client is None:
             client = self.get_client(write=True)
 
         key = self.make_key(key, version=version)
         return client.lock(
-            key, timeout=timeout, sleep=sleep, blocking_timeout=blocking_timeout
+            key,
+            timeout=timeout,
+            sleep=sleep,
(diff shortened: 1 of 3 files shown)


sindrig:
@niwinz is this project orphaned? Can we do something about that?
BTW looks like the build is failing because of https://github.com/niwinz/django-redis/pull/391

jdufresne:
Makes sense to me.

Could you add a test to demonstrate the new behavior?

WDYT about the suggestion in #427?

sindrig:
I added a test showing the usage and also forwarded the parameter to the sharded client.

I am not opposed to the general idea of having `**options` forwarded to the client, but it wouldn't be consistent with the api of this package. I feel like if we want to take that route we should introduce it to all wrapper functions at the same time.  In the end it's more of a design question IMO, and I guess it boils down to whether or not there are more cases like this one.

If you want to I can easily change my PR to do `**options` forwarding instead of `thread_local`.

jdufresne:
Thanks!

Issue #367, opened 2019-02-20
AssertionError: 'b' unexpectedly found in ['a', 'b']


jayvdb:
After fixing https://github.com/niwinz/django-redis/issues/366, I now [encounter on all Travis jobs](https://travis-ci.org/jayvdb/django-redis/builds/495996033). (And I also encounter this on my openSUSE builds, with [a workaround](https://build.opensuse.org/request/show/677659) )

```py
======================================================================
FAIL: test_keys (tests.test_backend.DjangoRedisCacheTestEscapePrefix)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/jayvdb/django-redis/tests/test_backend.py", line 104, in test_keys
    self.assertNotIn('b', keys)
AssertionError: 'b' unexpectedly found in ['a', 'b']

```

PR #391, opened 2019-09-14
Fix prefix escaping at ShardClient.keys function

Jamim:
Hello @niwinz,

This PR fixes #367 and compatibility with Django's `master` branch.

Currently, the `keys` function uses the `make_key` function which doesn't escape the prefix. In case the default `*` prefix is used and some other client uses the same redis db with non-default prefix, it misleads to fetching of prefixed keys of the other client.
In order to fix the issue, we have to use the `make_pattern` function instead of the `make_key` function.

Best regards!

PR #387, opened 2019-07-19
Drop support for Python 2

jdufresne:
Python 2 support is nearing end of life. At the end of this year, it will no longer be supported, including for security issues.

Django has already dropped support for Python 2 in Django 2.0. Follow upstream to be more compatible with the master branch as tests now pass.

For more details, see https://pythonclock.org/

All code changes and clean ups that were noticed were done. Changes include:

- Remove `str`/`unicode` compatibility shims.
- Remove encoding cookies from source files. Python 3 interprets all files as utf-8 by default.
- Remove unnecessary `__future__` imports. The features are now built-in.
- Use modern and simpler `super()` syntax.
- Update the wheel configuration to reflect it is no longer universal.
- Update trove classifiers.
- And more …

Also remove supported for end of life Python 3.4. It went end of life 2019-03-18.

PR #401, opened 2019-12-04
remove dependency on django.utils.six (for Django 3.0)

terencehonles:


terencehonles:
I added the CI fix from #391 which will close #367

jdufresne:
Is this a duplicate of #387 (has been merged) or is there more here?

terencehonles:
@jdufresne I didn't realize we'd drop Python 2 so soon. I think this is a duplicate, but I'm looking on mobile and I don't see something like https://github.com/niwinz/django-redis/commit/887204ec5915c4e52c2fe6476a5cfacd48110d6b in the diff for #387

terencehonles:
Nevermind it looks like you merged #391

dhcn:
When will this patch be released?

PR #402, opened 2019-12-04
Remove support for end-of-life Djangos

jdufresne:
Django 2.0 went EOL on April 1, 2019 and Django 2.1 on December 2, 2019.

These versions are no longer maintained. They are not receiving bug
fixes or even security fixes. Their use is not recommended.

Removing these EOL version will significantly reduce testing overhead.


terencehonles:
Should we make sure documentation mentions that we only support non EOL Django? I'm torn between suggesting the package support older versions a little longer vs encouraging people to make the jump. If I hadn't migrated our codebase to Python 3 last year we'd be hitting a lot of these issues 🤐. Since this PR is just editing the test matrix I'm not sure if it's really reducing the testing overhead. Maybe drop some of the redis versions?

Either way I'm fine however this goes. We're on the current release now 🙂

jdufresne:
> Should we make sure documentation mentions that we only support non EOL Django? 

Yeah definitely. The release notes will be sure to mention this.

> I'm not sure if it's really reducing the testing overhead.

Each test platform tested adds several minutes of test time. This is fine for supported platforms so we ensure the package works, but for unsupported platforms that projects should not be using, it is a lot of waiting.

PR #423, opened 2019-12-18
Fix DefaultClient.set method overriding empty Pipeline

dmarquezlutfy:
Issue:

`if not client` is always `True` if `client` is an empty Pipeline, so `set_many` operations are always executed one by one because `client` is overrode by `self.get_client`. 

Fix:

Change if condition to `if client is None`

Notes:

* Test for this has been included


jdufresne:
Thanks!

PR #451, opened 2020-05-13
Started typing codebase

WisdomPill:
I started giving types to the codebase using type hinting since python3.5 supports them and python3.5 is the minimum version required by the library.

I started with serializers and compressors.

I have some doubts about the client, specifically about key, version and prefix.

Redis supports everything that can be byte encoded like stated here https://redis.io/topics/data-types-intro

```
Redis keys are binary safe, this means that you can use any binary sequence as a key, from a string like "foo" to the content of a JPEG file. The empty string is also a valid key.
```

But other caches like `memcached` and django own `BaseCache` seem to point to key and prefix being a string and version an integer.

My question is, leaving Any instead of a more specific type or to enforce the types that I just mentioned?

By the way `django-stubs` here https://github.com/typeddjango/django-stubs/blob/master/django-stubs/core/cache/backends/base.pyi have specific types in some methods and on some not.


WisdomPill:
I ended up not using str for key and prefix and int for version since it was almost mandatory from the rest of the method calls.

I am in still in doubt of using str for key and prefix, I would prefer more feedback before taking conclusions

django_redis/serializers/pickle.py
```diff
@@ -1,24 +1,27 @@
...
 
 
 class PickleSerializer(BaseSerializer):
-    def __init__(self, options):
+    def __init__(self, options) -> None:
         self._pickle_version = -1
         self.setup_pickle_version(options)
 
-    def setup_pickle_version(self, options):
+        super(PickleSerializer, self).__init__(options=options)
```
jdufresne, line 14: As this code is Python 3 only, can use `super()` without arguments:

```suggestion
        super().__init__(options=options)
```

django_redis/client/default.py
```diff
@@ -15,16 +19,15 @@
 
 _main_exceptions = (TimeoutError, ResponseError, ConnectionError, socket.timeout)
 
-
 special_re = re.compile("([*?[])")
 
 
-def glob_escape(s):
+def glob_escape(s) -> Union[bytes, str]:
```
jdufresne, line 25: The regex is `str`, not `bytes`. Did you find it necessary to include `bytes`? If this is intended to support both, it is buggy and will produce a `BytesWarning`.

django_redis/client/default.py
```diff
@@ -108,14 +111,14 @@ def connect(self, index=0):
...
-        xx=False,
-    ):
+        key: Any,
+        value: Any,
+        timeout: Optional[float] = DEFAULT_TIMEOUT,
+        version: Optional[int] = None,
+        client: Redis = None,
+        nx: bool = False,
+        xx: bool = False,
+    ) -> Union[Optional[int], bool]:
```
jdufresne, line 121: When does this function return `int`? Or `None` for that matter.

django_redis/serializers/pickle.py
WisdomPill: Thanks, I did not know that

django_redis/client/default.py
WisdomPill: Yeah, I fixed it but did not push the modifications yet

WisdomPill: Sorry for that

django_redis/client/default.py
WisdomPill: the delete should return an `int`, actually it returns directly what the client returns, that should be `bytes`

WisdomPill:
I am testing locally with `mypy` I did not add it yet to the pipeline because I get a lot of errors as of now

jdufresne:
Thanks!

I think we should hold off on landing this change until mypy passes and the type checking can be added to CI. Without that, this adds a lot of noise and no guarantees of correctness. We can leave this PR open to track efforts in that direction.

WisdomPill:
Only one error is left.

```
django_redis/client/default.py:607: error: Item "str" of "Union[str, Any]" has no attribute "decode"
```

I cannot understand how to overcome this, it seems like the returning type is `Union[str, Any]`.

@jdufresne can you help with this?

WisdomPill:
Another thing is about `black` and `isort` it seems like they are in contrast

about imports in line 8 of `django_redis/client/default.py`

and imports in line 1 of `django_redis/compressors/lz4.py`

WisdomPill:
@jdufresne can you help with `black` and `isort` conflict?

jdufresne:
Try rebasing on 3fd04a378804272d71c651db62927e7beeb4c62e. That might fix the issue.

WisdomPill:
> Try rebasing on [3fd04a3](https://github.com/jazzband/django-redis/commit/3fd04a378804272d71c651db62927e7beeb4c62e). That might fix the issue.

Thanks it has fixed the issue.

Now the other clients are missing the type hints, I am going to work on it in the following days,
can you give a look to what's been done so far?

jdufresne:
> Ditched relative imports in client/default.py

Was there a particular reason for this? Why can't relative imports remain?

WisdomPill:
I find relative imports less explicit and prone to error, I am not strongly against them, but I avoid using them usually, I can restore them if necessary.

WisdomPill:
@jdufresne I added mypy to travis, can you check everything is okay?
I also restored relative imports.

terencehonles:
@WisdomPill did you want to update this to the latest master and I can review it after that? It would probably be good to get this into the codebase if there aren't too many errors from mypy.

WisdomPill:
Yes sure, thanks for the help, will do that this weekend

WisdomPill:
@terencehonles here you go, I fixed some type hints missing because of sentinel and added an ignore error for making flake8 and black compatible

terencehonles:
Cool, thanks. I'll try to look at this, this week.

django_redis/client/default.py
```diff
@@ -295,9 +342,7 @@ def delete_pattern(
 
         pattern = self.make_pattern(pattern, version=version, prefix=prefix)
 
-        kwargs = {"match": pattern}
-        if itersize:
-            kwargs["count"] = itersize
+        kwargs = {"match": pattern, "count": itersize}
```
terencehonles, line 345: These can probably be just inlined if they don't need this conditional checking (They don't on `master`, but not sure if that was the case before)

django_redis/client/default.py
```diff
@@ -284,8 +326,13 @@ def delete(self, key, version=None, prefix=None, client=None):
...
 
     def delete_pattern(
-        self, pattern, version=None, prefix=None, client=None, itersize=None
-    ):
+        self,
+        pattern: str,
+        version: Optional[int] = None,
+        prefix: Optional[str] = None,
+        client: Redis = None,
+        itersize=None,
```
terencehonles, line 334:
```suggestion
        client: Optional[Redis] = None,
        itersize: Optional[int] = None,
```

django_redis/client/default.py
```diff
@@ -271,7 +307,13 @@ def lock(
             thread_local=thread_local,
         )
 
-    def delete(self, key, version=None, prefix=None, client=None):
+    def delete(
+        self,
+        key: Any,
+        version: Optional[int] = None,
+        prefix: Optional[str] = None,
+        client: Redis = None,
```
terencehonles, line 315:
```suggestion
        client: Optional[Redis] = None,
```

django_redis/client/default.py
```diff
@@ -252,11 +288,11 @@ def expire(self, key, timeout, version=None, client=None):
     def lock(
         self,
         key,
-        version=None,
+        version: Optional[int] = None,
         timeout=None,
         sleep=0.1,
         blocking_timeout=None,
-        client=None,
+        client: Redis = None,
```
terencehonles, line 295:
```suggestion
        client: Optional[Redis] = None,
```

django_redis/client/default.py
```diff
@@ -231,7 +263,9 @@ def get(self, key, default=None, version=None, client=None):
 
         return self.decode(value)
 
-    def persist(self, key, version=None, client=None):
+    def persist(
+        self, key: Any, version: Optional[int] = None, client: Redis = None
```
terencehonles, line 267:
```suggestion
        self, key: Any, version: Optional[int] = None, client: Optional[Redis] = None
```

django_redis/client/default.py
```diff
@@ -202,15 +221,28 @@ def incr_version(self, key, delta=1, version=None, client=None):
...
         """
         return self.set(key, value, timeout, version=version, client=client, nx=True)
 
-    def get(self, key, default=None, version=None, client=None):
+    def get(
+        self,
+        key: Any,
+        default=None,
+        version: Optional[int] = None,
+        client: Redis = None,
```
terencehonles, line 244:
```suggestion
        client: Optional[Redis] = None,
```

django_redis/client/default.py
```diff
@@ -202,15 +221,28 @@ def incr_version(self, key, delta=1, version=None, client=None):
...
 
         Returns ``True`` if the object was added, ``False`` if not.
         """
         return self.set(key, value, timeout, version=version, client=client, nx=True)
 
-    def get(self, key, default=None, version=None, client=None):
+    def get(
+        self,
+        key: Any,
+        default=None,
```
terencehonles, line 242:
```suggestion
        default: Optional[Any] = None,
```

django_redis/client/default.py
```diff
@@ -202,15 +221,28 @@ def incr_version(self, key, delta=1, version=None, client=None):
...
         return version + delta
 
-    def add(self, key, value, timeout=DEFAULT_TIMEOUT, version=None, client=None):
+    def add(
+        self,
+        key: Any,
+        value: Any,
+        timeout: Any = DEFAULT_TIMEOUT,
+        version: Optional[Any] = None,
+        client: Redis = None,
```
terencehonles, line 230:
```suggestion
        client: Optional[Redis] = None,
```

django_redis/client/default.py
```diff
@@ -202,15 +221,28 @@ def incr_version(self, key, delta=1, version=None, client=None):
         self.delete(old_key, client=client)
         return version + delta
 
-    def add(self, key, value, timeout=DEFAULT_TIMEOUT, version=None, client=None):
+    def add(
+        self,
+        key: Any,
+        value: Any,
+        timeout: Any = DEFAULT_TIMEOUT,
+        version: Optional[Any] = None,
```
terencehonles, line 229: Should these not be constrained more?

django_redis/client/default.py
```diff
@@ -170,7 +183,13 @@ def set(
                     continue
                 raise ConnectionInterrupted(connection=client) from e
 
-    def incr_version(self, key, delta=1, version=None, client=None):
+    def incr_version(
+        self,
+        key: Any,
+        delta: int = 1,
+        version: Optional[int] = None,
+        client: Redis = None,
```
terencehonles, line 191:
```suggestion
        client: Optional[Redis] = None,
```

django_redis/client/default.py
```diff
@@ -113,14 +124,14 @@ def disconnect(self, index=0, client=None):
...
-        version=None,
-        client=None,
-        nx=False,
-        xx=False,
-    ):
+        key: Any,
+        value: Any,
+        timeout: Optional[float] = DEFAULT_TIMEOUT,
+        version: Optional[int] = None,
+        client: Redis = None,
```
terencehonles, line 131:
```suggestion
        client: Redis = Optional[None],
```

django_redis/client/default.py
```diff
@@ -59,17 +60,22 @@ def __init__(self, server, params, backend):
 
         self.connection_factory = pool.get_connection_factory(options=self._options)
 
-    def __contains__(self, key):
+    def __contains__(self, key: Any) -> bool:
         return self.has_key(key)
 
-    def get_next_client_index(self, write=True, tried=()):
+    def get_next_client_index(
+        self, write: bool = True, tried: Optional[List[int]] = None
```
terencehonles, line 67: Should this not be widened to instead?:

```suggestion
        self, write: bool = True, tried: Optional[Collection[int]] = ()
```

terencehonles (commented):
I've not had a change to run `mypy` but here are some suggestions based on the diff.

django_redis/client/default.py
WisdomPill: I think there is a little typo here, but I understood, it is in line with the other suggestions

django_redis/client/default.py
WisdomPill: Well, yes and no. Technically key and value could be `ByteStr`, I was not sure because you could decide to use many types and they will be valid for redis. What do you suggest? I would keep it this way and discuss about it elsewhere, at the time I even asked on django stubs repository because it was not typed even there, don't know now, I will check

django_redis/hash_ring.py
```diff
@@ -28,8 +29,8 @@ def add_node(self, node):
     def remove_node(self, node):
         self.nodes.remove(node)
         for x in range(self.replicas):
-            _hash = hashlib.sha256("%s:%d" % (node, x)).hexdigest()
-            self.ring.remove(_hash)
+            _hash = hashlib.sha256(("%s:%d" % (node, x)).encode()).hexdigest()
```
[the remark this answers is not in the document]
WisdomPill: Sure

django_redis/pool.py
```diff
@@ -1,10 +1,12 @@
+from typing import Dict
 from urllib.parse import parse_qs, urlparse
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.module_loading import import_string
+from redis import Redis
 from redis.connection import DefaultParser, to_bool
-from redis.sentinel import Sentinel
+from redis.sentinel import Sentinel  # type: ignore
```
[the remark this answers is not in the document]
WisdomPill: I remember I did try but it did not work, will try again

django_redis/pool.py
```diff
@@ -61,7 +63,7 @@ def make_connection_params(self, url):
 
         return kwargs
 
-    def connect(self, url):
+    def connect(self, url: str) -> Redis:
```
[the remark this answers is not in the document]
WisdomPill: Could you elaborate more?

django_redis/client/default.py
terencehonles: lol, yeah, I should have just marked one and left you to `s/Redis/Optional[Redis]/` :sweat_smile:

django_redis/client/default.py
terencehonles: I see, well I'm not sure why things are loosely defined on django-stubs, and I'm fine with leaving `Any`, but not sure if that's much better than leaving it undefined :confused:

django_redis/pool.py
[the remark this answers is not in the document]
terencehonles: I'll try to checkout the branch and run mypy myself soon.

django_redis/pool.py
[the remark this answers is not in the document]
terencehonles: Something like:

```python
from typing import Generic, TypeVar

PoolType = TypeVar('PoolType', covariant=True, bound=redis.connection.ConnectionPool)
RedisType = TypeVar('PoolType', covariant=True, bound=redis.client.Redis)

class ConnectionFactory(Generic[PoolType, RedisType]):
    def connect(self, url: str) -> RedisType: ...
    def get_or_create_connection_pool(self, params: Mapping[str, Any]) -> PoolType: ...
    def get_connection_pool(self, params: Mapping[str, Any]) -> PoolType: ...
```

which would allow someone to either write a mypy plugin or explicitly declare what types fill in the generic types. However, I'm not sure if people would be likely declaring a typed ConnectionFactory since the types will come from the options in `__init__`.

django_redis/pool.py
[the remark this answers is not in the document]
WisdomPill: fixed

django_redis/client/default.py
WisdomPill: It should always be a list, if I am not mistaken I purposely changed it because before it could have been tuple and list

django_redis/client/default.py
WisdomPill: I see you are a `sed` user. Nice to meet you mate

django_redis/compressors/lz4.py
```diff
@@ -1,5 +1,5 @@
-from lz4.frame import compress as _compress
-from lz4.frame import decompress as _decompress
+from lz4.frame import compress as _compress  # type: ignore
```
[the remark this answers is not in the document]
WisdomPill: fixed

django_redis/client/sentinel.py
```diff
@@ -1,7 +1,7 @@
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 from django.core.exceptions import ImproperlyConfigured
-from redis.sentinel import SentinelConnectionPool
+from redis.sentinel import SentinelConnectionPool  # type: ignore
```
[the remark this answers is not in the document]
WisdomPill: fixed

django_redis/client/default.py
```diff
@@ -567,9 +656,9 @@ def make_pattern(self, pattern, version=None, prefix=None):
 
         if version is None:
             version = self._backend.version
-        version = glob_escape(str(version))
+        version_str = glob_escape(str(version))
```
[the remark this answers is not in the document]
WisdomPill: it is already like this, am I wrong?

django_redis/client/default.py
```diff
@@ -515,7 +592,13 @@ def has_key(self, key, version=None, client=None):
         except _main_exceptions as e:
             raise ConnectionInterrupted(connection=client) from e
 
-    def iter_keys(self, search, itersize=None, client=None, version=None):
+    def iter_keys(
+        self,
+        search: str,
+        itersize=None,
+        client: Redis = None,
```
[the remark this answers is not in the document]
WisdomPill: done

django_redis/client/default.py
```diff
@@ -470,14 +537,22 @@ def incr(self, key, delta=1, version=None, client=None, ignore_key_check=False):
...
+    ) -> int:
         """
         Decreace delta to value in the cache. If the key does not exist, raise a
         ValueError exception.
         """
         return self._incr(key=key, delta=-delta, version=version, client=client)
 
-    def ttl(self, key, version=None, client=None):
+    def ttl(
+        self, key: Any, version: Optional[int] = None, client: Redis = None
```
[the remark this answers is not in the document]
WisdomPill: done

django_redis/client/default.py
```diff
@@ -456,7 +516,14 @@ def _incr(self, key, delta=1, version=None, client=None, ignore_key_check=False)
 
         return value
 
-    def incr(self, key, delta=1, version=None, client=None, ignore_key_check=False):
+    def incr(
+        self,
+        key: Any,
+        delta: int = 1,
+        version: Optional[int] = None,
+        client: Redis = None,
```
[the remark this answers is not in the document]
WisdomPill: done

django_redis/client/default.py
```diff
@@ -470,14 +537,22 @@ def incr(self, key, delta=1, version=None, client=None, ignore_key_check=False):
             ignore_key_check=ignore_key_check,
         )
 
-    def decr(self, key, delta=1, version=None, client=None):
+    def decr(
+        self,
+        key: Any,
+        delta: int = 1,
+        version: Optional[int] = None,
+        client: Redis = None,
```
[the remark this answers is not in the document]
WisdomPill: done

django_redis/client/default.py
```diff
@@ -366,7 +411,9 @@ def encode(self, value):
 
         return value
 
-    def get_many(self, keys, version=None, client=None):
+    def get_many(
+        self, keys, version: Optional[int] = None, client: Redis = None
```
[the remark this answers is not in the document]
WisdomPill: done

django_redis/client/default.py
```diff
@@ -326,7 +371,7 @@ def delete_many(self, keys, version=None, client=None):
         except _main_exceptions as e:
             raise ConnectionInterrupted(connection=client) from e
 
-    def clear(self, client=None):
+    def clear(self, client: Redis = None) -> None:
```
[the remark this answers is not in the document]
WisdomPill: done

django_redis/client/default.py
```diff
@@ -308,7 +353,7 @@ def delete_pattern(
         except _main_exceptions as e:
             raise ConnectionInterrupted(connection=client) from e
 
-    def delete_many(self, keys, version=None, client=None):
+    def delete_many(self, keys, version: Optional[int] = None, client: Redis = None):
```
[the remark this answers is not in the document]
WisdomPill: done

django_redis/client/default.py
WisdomPill: done

django_redis/client/default.py
WisdomPill: done

django_redis/client/default.py
WisdomPill: done

django_redis/client/default.py
WisdomPill: I don't see it anymore

django_redis/client/default.py
WisdomPill: done

django_redis/client/default.py
WisdomPill: done

setup.cfg
```diff
@@ -51,10 +51,12 @@ skip_covered = true
 ignore =
     W503
     W601
+    E203
```
terencehonles, line 54: Do we need this? I tried removing it and it didn't seem to trigger any warnings.

django_redis/hash_ring.py
```diff
@@ -25,33 +28,35 @@ def add_node(self, node):
...
         _hash = hashlib.sha256(key.encode()).hexdigest()
         idx = bisect.bisect(self.sorted_keys, _hash)
         idx = min(idx - 1, (self.replicas * len(self.nodes)) - 1)
-        return (self.ring[self.sorted_keys[idx]], idx)
+        return self.ring[self.sorted_keys[idx]], idx
 
-    def iter_nodes(self, key):
+    def iter_nodes(
+        self, key: str
+    ) -> Generator[Tuple[Optional[str], Optional[str]], str, None]:
```
terencehonles, line 53: This is probably the more compact form (see the last example here: https://docs.python.org/3/library/typing.html#typing.Generator)

```suggestion
    ) -> Iterator[Tuple[Optional[str], Optional[str]]]:
```

django_redis/hash_ring.py
```diff
@@ -1,19 +1,22 @@
...
 class HashRing:
-    nodes = []
+    nodes = []  # type: List[str]
 
-    def __init__(self, nodes=(), replicas=128):
-        self.replicas = replicas
-        self.ring = {}
-        self.sorted_keys = []
+    def __init__(
+        self, nodes: Union[List[str], Tuple[str, ...]] = (), replicas: int = 128
```
terencehonles, line 10: This is probably more appropriate because the order is respected but you can pass any iterable (or `Sequence[str]`).

```suggestion
        self, nodes: Iterable[str] = (), replicas: int = 128
```

django_redis/client/default.py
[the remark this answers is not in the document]
terencehonles: I'm suggesting removing the variable and putting `glob_escape(...)` in the key_func call below.

terencehonles (approved):
LGTM, a few comments, but it looks like you spent some time adding more types so I'm happy with the increased coverage.

terencehonles:
Do you think this has sufficient typing coverage to expose it to others or do you want to add more before doing so?

I added mypy to the coverage information, but I'm not sure how you can view it separately. I figured I'd let you know about the `--html-report` option and since I noticed that codecov.io used a cobertura report and I knew that mypy could generate that I figured that might make it easier to answer that question. However, I'm not sure if it really does and what the coverage report name actually does.

terencehonles:
Looks like the documentation suggests flags are what I should have been using not the name.

terencehonles:
So that last commit doesn't look like it was completely required. It just will show separate checks for mypy and the tests in the PR checks section. I think it's fine to leave it as it was, but what I was trying to figure out was how to see the coverage for mypy vs the tests in their UI. I had found an example repo that did what I had wanted, but it looks like the real difference was I was looking at the commit rather than the PR view. If you look at the commit view it will show the flags above the diff and you can toggle which coverage report you are looking at based on the flag that it was uploaded for.

For example, on the UI https://codecov.io/gh/jazzband/django-redis/src/c6a79d70ffa5816ac045cb26441fc6eeb0820031/django_redis/cache.py 
Screenshot from 2021-04-11 13-00-47 the two flags on the top right are toggles to pick which reports to show. I'm not sure why they don't enable that on the PR view, but just a FYI in case you were trying to look at the coverage information without generating the HTML report manually locally.

setup.cfg
terencehonles: I see this is from https://black.readthedocs.io/en/stable/compatible_configs.html#id1

django_redis/hash_ring.py
WisdomPill: But it has a `SendType` so Iterator is not really describing the `Generator`.
It is arguable that default to `None` is not a good returning type, typingwise

django_redis/hash_ring.py
WisdomPill: I did not know `Iterable` could be used for tuple as well. I usually prefer to have a more specific type but I can't argue with simplicity

WisdomPill:
So I can add the latest changes and then we can get this merged?

After that we can increase coverage even more with some good first issues for the young souls of the gituniverse?

django_redis/hash_ring.py
terencehonles: But the send wouldn't be used here, right? I'm aware of the difference I just don't think it matters for this code

WisdomPill:
@terencehonles can you give this a quick review. I think I did what was supposed to be done, but I might be wrong, the changes are too many to keep track of everything.

django_redis/hash_ring.py
```diff
@@ -1,19 +1,20 @@
...
+    def __init__(self, nodes: Iterable[str] = (), replicas: int = 128):
+        self.replicas: int = replicas
+        self.ring: Dict[str, str] = {}
+        self.sorted_keys: List[str] = []
 
         for node in nodes:
             self.add_node(node)
 
-    def add_node(self, node):
+    def add_node(self, node: str):
```
terencehonles, line 17: I don't think this is automatically implied so you might as well add it to this and the others

```suggestion
    def add_node(self, node: str) -> None:
```

terencehonles (approved):
I only saw a few changes since my last review so just starting from there. I noticed something that could be adjusted, but it can also come in a follow up

django_redis/hash_ring.py
WisdomPill: Done

WisdomPill:
@terencehonles friendly ping for the hopefully final review

terencehonles:
> @terencehonles friendly ping for the hopefully final review

Lol, I was just waiting for you to merge since the first approval :joy: and the second was fine too that's why I had said the small suggestion could be made in a separate PR.

It looked good to me, but before we declare we're typed to external packages I'd like to get some more types in the system, but this is a good first step towards that.

WisdomPill:
So for now we can merge it, not declare that the library is typed and put out some `good first issue` to finish typing?

terencehonles:
Yeah, I'm good with that. Not sure if we _need_ to create an issue, since I don't think I would expect to have someone try to finish typing in one go. I would expect over time our coverage would get to a point where we're happy with declaring the package as typed and we'll enable that then. How we want to get to that is not _super_ important, but if you want to create an issue or issues for subparts of the project then I'm fine with that.

PR #458, opened 2020-07-14
add parallel testing example to README

mcleanmds:
Based of recent experience getting Django parallel testing working with a redis db, it may be useful to document the approach.

Happy for any feedback/changes.

If this is outside the scope of what the README should cover, please reject.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -149,6 +149,62 @@ following lines to your test class:
     def tearDown(self):
         get_redis_connection("default").flushall()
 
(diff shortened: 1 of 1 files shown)


WisdomPill:
Hello @mcleanmds!

First and foremost, thanks for the contribution.
Can you provide a test case for this?

I would also point to those tests in the docs.

This is mainly to keep track of any api changes that might happen in django in the future

WisdomPill (changes requested):
Add it to the test suite please

Issue #360, opened 2019-01-17
Expireat support


cashaev:
Dear team!

Support for EXPIREAT command would be much appreciated!

WisdomPill:
Hello @cashaev!

I think it is quite easy to implement, do you fancy giving it a shot and open a PR?

cashaev:
> Hello @cashaev!
> 
> I think it is quite easy to implement, do you fancy giving it a shot and open a PR?

Hi, @WisdomPill !
Thank you for the offer. Though it is quite late, but I will take a chance and contribute with pleasure.

Please, check the PR [https://github.com/jazzband/django-redis/pull/472](url)

WisdomPill:
Thanks for the contribution @cashaev !

Issue #471, opened 2020-10-09
Job failing on Travis-CI for Intel


kishorkunal-raj:
Hi All,
I am trying to run the jobs on Travis CI and been observed that job for TOXENV=lint  on AMD64 if failing. Logs are mentioned below

$ python --version

Python 3.6.7

$ pip --version

pip 20.1.1 from /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pip (python 3.6)
install

1.76s$ pip install tox

8.09s$ tox

lint create: /home/travis/build/kishorkunal-raj/django-redis/.tox/lint

lint installdeps: black, flake8, isort >= 5.0.2

lint installed: appdirs==1.4.4,black==20.8b1,click==7.1.2,dataclasses==0.7,flake8==3.8.4,importlib-metadata==2.0.0,isort==5.6.1,mccabe==0.6.1,mypy-extensions==0.4.3,pathspec==0.8.0,pycodestyle==2.6.0,pyflakes==2.2.0,regex==2020.9.27,toml==0.10.1,typed-ast==1.4.1,typing-extensions==3.7.4.3,zipp==3.3.0

lint run-test-pre: PYTHONHASHSEED='67920843'

lint run-test: commands[0] | black --target-version py35 --check --diff .

--- tests/test_backend.py	2020-10-09 12:29:27.773883 +0000

+++ tests/test_backend.py	2020-10-09 12:29:42.340158 +0000

@@ -321,17 +321,22 @@

         key = "key"

         value = "value"

 

         with patch.object(pipeline, "set") as mocked_set:

             self.cache.set(

-                key, value, client=pipeline,

+                key,

+                value,

+                client=pipeline,

             )

 

         if isinstance(self.cache.client, herd.HerdClient):

             default_timeout = self.cache.client._backend.default_timeout

             herd_timeout = (default_timeout + herd.CACHE_HERD_TIMEOUT) * 1000

-            herd_pack_value = self.cache.client._pack(value, default_timeout,)

+            herd_pack_value = self.cache.client._pack(

+                value,

+                default_timeout,

+            )

             mocked_set.assert_called_once_with(

                 self.cache.client.make_key(key, version=None),

                 self.cache.client.encode(herd_pack_value),

                 nx=False,

                 px=herd_timeout,

would reformat tests/test_backend.py

Oh no! 💥 💔 💥

1 file would be reformatted, 41 files would be left unchanged.

[...]

PR #472, opened 2020-10-09
add expire_at method

cashaev:
Adding support for EXPIREAT key timestamp method in DefaultClient and ShardedClient.

fixes: #360


WisdomPill:
Looks okay, 
I think that this method and some others need to give back the result of the operation (a boolean).

It is my first PR review, so I ask @jdufresne to confirm if everything looks okay.

cashaev:
> Looks okay,
> I think that this method and some others need to give back the result of the operation (a boolean).
> 
> It is my first PR review, so I ask @jdufresne to confirm if everything looks okay.

Maybe so. But adding the result should definitely be consistent across all the methods, and possibly done in a separate PR?
Otherwise, this PR's scope will expand too much.
@WisdomPill Thank you very much for a prompt reply!

WisdomPill (commented):
Thanks for solving #471 as well

WisdomPill (approved):
Looks good to me

WisdomPill:
Hello @cashaev can you rebase on master so that we can merge this?

Thanks

cashaev:
Any action on this PR? I need this functionality again now )))

cashaev:
> Hello @cashaev can you rebase on master so that we can merge this?
> 
> Thanks

let me try

cashaev:
> Hello @cashaev can you rebase on master so that we can merge this?
> 
> Thanks

Sorry, not sure if I can do it correctly. Any assistance? Do I need to reopen it from my master branch?

jdufresne:
Please rebase. This should not contain changes to `.github/workflows/`, but they current show up in the diff.

WisdomPill:
@cashaev let me explain the steps to rebase on `django-redis/master`.

- locally add another remote pointing to `jazzband/django-redis` with the following command: `git remote add upstream git@github.com:jazzband/django-redis.git`
- fetch from remote master `git fetch upstream master`
- make sure to be on your master with the following command: `git checkout master`. Or just check with `git status`
- rebase your master (`origin/master`) on jazzband's master with the following command: `git rebase upstream/master`
- rebase your feature branch on your own master with the following two commands `git checkout expireat` and `git rebase origin/master`
- if there will be a need to merge and stuff you will get prompts, after that you can push with `git push --force origin expireat`, force is needed because you are changing the history of you branch.

Theoretically you could skip rebasing your master on upstream/master, but it is just good practice to keep everything in sync.

Hope this helps, let me know if you need more assistance

terencehonles (approved):
LGTM, but I'm not sure if we want to be implementing _every_ redis command in this class. @WisdomPill thoughts? You want to squash this if you're good with it so we can include it in the next release?

WisdomPill:
I think it is just a utility method recycling `.expire` method so it is okay to add it, I am worried that many commands are implemented only in `DefaultClient` so I think other clients could use some enhancements to be able to do the same things as `DefaultClient`.

tests/test_backend.py
```diff
@@ -661,6 +661,31 @@ def test_expire(self):
...
+        self.assertAlmostEqual(ttl, timedelta(hours=1).total_seconds(), delta=1)
+
+        # Test settings expiration time 1 hour ahead by Unix timestamp.
+        self.cache.set("foo", "bar", timeout=None)
+        expiration_time = datetime.datetime.now() + timedelta(hours=2)
+        self.cache.expire_at("foo", int(expiration_time.timestamp()))
+        ttl = self.cache.ttl("foo")
+        self.assertAlmostEqual(ttl, timedelta(hours=1).total_seconds() * 2, delta=1)
+
+        """
```
WisdomPill, line 680: Can this be a multiline comment?

WisdomPill, line 680:
```suggestion
        # Test settings expiration time 1 hour in past, which
        # effectively deletes the key.
        
```

django_redis/client/default.py
```diff
@@ -289,6 +290,19 @@ def expire(
         if client.exists(key):
             client.expire(key, timeout)
 
+    def expire_at(self, key, when: Union[datetime, int], version=None, client=None):
```
WisdomPill, line 293:
```suggestion
    def expire_at(self, key: Any, when: Union[datetime, int], version: Optional[int]=None, client: Optional[Redis]=None):
```

WisdomPill (changes requested):

terencehonles:
@WisdomPill I just applied your suggestions myself. If you're good with it as is, I'll let you handle squashing this.

WisdomPill (approved):

WisdomPill:
Thanks @terencehonles!

WisdomPill:
wait a mement

WisdomPill (changes requested):
`when` can be a `datetime` but using a date time will get an exception, `datetime` needs to be converted to unix timestamp

terencehonles:
> `when` can be a `datetime` but using a date time will get an exception, `datetime` needs to be converted to unix timestamp

Where? The tests cover sending both a `datetime` and an `int`, so that's covered there. The underlying client (if `redis-py`) handles converting to an `int` https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/client.py#L1659-L1660

WisdomPill:
Okay, sorry did not notice that redis py supports it, my bad

PR #475, opened 2020-10-14
Fix for lint issue

kishorkunal-raj:
Hi,
Job (TOXENV=lint) was failing for Intel architecture on travis-ci. I had added the fix for the issue in PR.
Kindly review and merge if found relevant

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -323,13 +323,18 @@ class DjangoRedisCacheTests(unittest.TestCase):
 
         with patch.object(pipeline, "set") as mocked_set:
             self.cache.set(
-                key, value, client=pipeline,
+                key,
+                value,
+                client=pipeline,
             )
 
         if isinstance(self.cache.client, herd.HerdClient):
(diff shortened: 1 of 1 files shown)


tests/test_backend.py
```diff
@@ -323,13 +323,18 @@ def test_set_call_empty_pipeline(self):
 
         with patch.object(pipeline, "set") as mocked_set:
             self.cache.set(
-                key, value, client=pipeline,
+                key,
+                value,
+                client=pipeline,
```
jdufresne, line 328: Removing the last comma will let Black reformat the statement to a single line rather than forcing this to wrap.

kishorkunal-raj: I had raised an issue regarding failure of job https://github.com/jazzband/django-redis/issues/471#issuecomment-707047419 and I got the above from there. 
So any further suggestions in order to avoid this?

jdufresne:
Merged as dbb90a2a127bf394972fb1aa4bfb7cabd03ced3f. Thanks for the report and the PR.

PR #476, opened 2020-10-14
Adding ppc64le architecture support on travis-ci

kishorkunal-raj:
Hi,
I had added ppc64le(Linux on Power) architecture support on travis-ci in the branch and looks like its been successfully added. I believe it is ready for the final review and merge. The travis ci build logs can be verified from the link below.
https://travis-ci.com/github/kishorkunal-raj/django-redis/builds/189878828

Reason behind running tests on ppc64le: This package is included in the ppc64le versions of RHEL and Ubuntu - this allows the top of tree to be tested continuously as it is for Intel, making it easier to catch any possible regressions on ppc64le before the distros begin their clones and builds. This reduces the work in integrating this package into future versions of RHEL/Ubuntu.

Please have a look.

Regards,
Kishor Kunal Raj

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,6 +56,98 @@ matrix:
       env: TOXENV=py38-djangomaster-redislatest
     - python: 3.8
       env: TOXENV=py38-djangomaster-redismaster
+    - python: 3.9
+      env: TOXENV=py39-djangomaster-redislatest
+    - python: 3.9
+      env: TOXENV=py39-djangomaster-redismaster
+      # Adding jobs for ppc64le architecture
+    - env: TOXENV=lint
+      arch: ppc64le
+    - python: 3.5
+      arch: ppc64le
+      env: TOXENV=py35-django22-redislatest
+    - python: 3.5
+      arch: ppc64le
+      env: TOXENV=py35-django22-redismaster
+    - python: 3.6
+      arch: ppc64le
+      env: TOXENV=py36-django22-redislatest
+    - python: 3.6
+      arch: ppc64le
+      env: TOXENV=py36-django22-redismaster
+    - python: 3.7
+      arch: ppc64le
+      env: TOXENV=py37-django22-redislatest
+    - python: 3.7
+      arch: ppc64le
+      env: TOXENV=py37-django22-redismaster
+    - python: 3.8
+      arch: ppc64le
+      env: TOXENV=py38-django22-redislatest
+    - python: 3.8
+      arch: ppc64le
+      env: TOXENV=py38-django22-redismaster
+    - python: 3.6
+      arch: ppc64le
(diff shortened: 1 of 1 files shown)


WisdomPill:
I do not know anything about `ppcle64`, for that matter if this architecture should be added then also the others should for example arm that probably is used more than `ppcle64`. 
Furthermore, if I am not mistaken, there is no platform dependent code. 

I do not think this is useful, not even django is tested with different architectures on its CI.

I leave the PR here, for others to review and merge if they think it is useful.

gerrith3:
@WisdomPill this package, as @kishorkunal-raj is included in debian and Ubuntu ppc64le distros, so it is at least valued by those communities.  I can't speak to how often it is used, although django and redis are both pretty common among our customers and even redis in some of our IBM products. As far as how often we find regressions - it happens more often across python projects than you might expect - often because of something like pulling in or expecting some intel based executable or rarely trying to access something that isn't pre-installed on a system or pre-built for a distro.  The good thing here is that we provide the Power/ppc64le resources for free and encourage communities to run their regressions tests to ensure there isn't an issue.  This simplifies things downstream for anyone who picks up any project and can know with confidence that they'll work on their big expensive hardware.  ;)

gerrith3:
@WisdomPill actually, IBM builds a container of django on Power (possibly my team does that), but so does Anaconda, possibly conda-forge, Alpine, SuSE, Fedora/Red Hat, Debian/Ubuntu, etc.  We are trying to consolidate more of these builds to the individual projects so we don't have 10 different sources reporting and trying to fix bugs in 20,000 communities when we go to build and update each distro.  So, we would definitely appreciate it if this build could be added, simplifying things downstream for no additional real cost here.  And, since it is intended to be architecture independent, it should add very little development overhead (builds are done in parallel, so no net add to the build times in most cases).  BTW, it looks like @kishorkunal-raj may need to update the patch now since there seem to be some conflicts in .travis.yml.

kishorkunal-raj:
@WisdomPill I had resolved the conflict for .travis.yml. I am not sure why I am not able to mark it as "resolve conflicts". Please have a look

WisdomPill:
@kishorkunal-raj you need to rebase your changes on the master branch of jazzband repo.

I repeat, I do not see why we should add this to the builds, to be honest `arm` should be added first if any architectures should be added. And no, it does add overhead to the build time, only five jobs at a time are possibile for open source software. 
Furthermore, why `django` does not do multiple architecture in its continuous integration pipelines? If django does not have why should we? 
Django is a dependency of a lot of other packages where you want to add builds for your architecture, but why not add them there, I am sure there is a higher probability of having some platform dependent code there than here.

jezdez:
This won't happen, see https://github.com/jazzband/django-model-utils/pull/448 for more details.

Issue #474, opened 2020-10-13
Add python3.9 to CI


WisdomPill:
Travis should add python 3.9 to the python versions for testing, maybe at the beginning we can leave it on the allowed failures if it causes issues

peymanslh:
Can I work on this issue?

WisdomPill:
Yes, thanks for the PR

PR #477, opened 2020-10-14
Add Python 3.9 to CI

peymanslh:
Changes:
- Add Python 3.9 to TravisCI config file

Closes: #474

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ matrix:
       env: TOXENV=py38-django22-redislatest
     - python: 3.8
       env: TOXENV=py38-django22-redismaster
+    - python: "3.9-dev"
+      env: TOXENV=py39-django22-redislatest
+    - python: "3.9-dev"
+      env: TOXENV=py39-django22-redismaster
     - python: 3.6
       env: TOXENV=py36-django30-redislatest
     - python: 3.6
@@ -28,6 +32,10 @@ matrix:
       env: TOXENV=py38-django30-redislatest
     - python: 3.8
       env: TOXENV=py38-django30-redismaster
+    - python: "3.9-dev"
+      env: TOXENV=py39-django30-redislatest
+    - python: "3.9-dev"
+      env: TOXENV=py39-django30-redismaster
     - python: 3.6
       env: TOXENV=py36-django31-redislatest
     - python: 3.6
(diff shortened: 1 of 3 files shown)


WisdomPill:
There is no python 3.9 yet on travis, python3.9 has been released just 10 days ago and travis still needs to support it.

They only support python3.9-dev as of now.

Please switch to python3.9-dev for this moment and leave it on the allowed to fail builds

WisdomPill (changes requested):
use py39-dev because py39 is not yet available

jdufresne:
What is the reason only "djangomaster" is being tested by Travis? Why not all Django targets?

peymanslh:
> What is the reason only "djangomaster" is being tested by Travis? Why not all Django targets?

I saw Python 3.8 was only tested with `django master`.
We have `django22`, `django30`, `django31`, and `djangomaster` in `tox.ini`.
Should I add all of these for Python `3.9-dev`?

WisdomPill:
> I checked Python 3.8 and it was only tested with django master.

Where have you seen that?

hugovk:
How about adding the Trove classifier to `setup.cfg` too?

```
    Programming Language :: Python :: 3.9
```

.travis.yml
```diff
@@ -56,13 +68,25 @@ matrix:
...
     - env: TOXENV=py37-djangomaster-redislatest
     - env: TOXENV=py37-djangomaster-redismaster
     - env: TOXENV=py38-djangomaster-redislatest
     - env: TOXENV=py38-djangomaster-redismaster
+    - env: TOXENV=py39-dev-django22-redislatest
+    - env: TOXENV=py39-dev-django22-redismaster
+    - env: TOXENV=py39-dev-django30-redislatest
+    - env: TOXENV=py39-dev-django30-redismaster
+    - env: TOXENV=py39-dev-django31-redislatest
+    - env: TOXENV=py39-dev-django31-redismaster
```
jdufresne, line 87: Thanks for keeping at it!

```suggestion
```

IMO, only `djangomaster` should allow failures, as it is unstable and changing. Python 3.9 is stable and no new PRs should break it.

peymanslh:
@jdufresne Thanks for your review.
I've fixed the conflict with the last commit.

jdufresne (approved):
LGTM. Thanks!

@WisdomPill Let me know if you'd like to a take a last look.

WisdomPill (approved):
Nice job, thanks to everybody who contributed

PR #478, opened 2020-10-15
Drop support for EOL Python 3.5

hugovk:
EOL at the end of last month: https://devguide.python.org/devcycle/#end-of-life-branches

Helps free up a couple of slots in a busy CI, and allows to use new features of Python (hello f-strings!).

And upgrade Python syntax with https://github.com/asottile/pyupgrade using `--py36-plus`.


jdufresne (approved):
Thank you

PR #481, opened 2020-11-08
Switch CI configuration to GitHub actions

jdufresne:


WisdomPill:
Hello @jdufresne, just out of curiosity, why the change to GitHub actions?

jdufresne:
As of late, Travis has been responding extremely slowly making it difficult to get for contributors to get timely feedback. GH actions doesn't have this same issue.

Another thing I like better, GH Actions show each combo of the matrix as a different integration on the PR. When one combo of the matrix fails in Travis, it makes it look like the entire test suite failed, but not so for GH actions.

WisdomPill:
It looks to me that also GitHub actions offers more concurrent jobs than travis, it will be a huge performance improvement

PR #482, opened 2020-11-13
Specified ACL login usage in the README

GrazingScientist:
It took me some digging in the redis-py source code (their README does not even hint this feature). Hence, I thought it may be useful to have it explicitely in the documentation for others

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -85,9 +85,19 @@ django-redis uses the redis-py native URL notation for connection strings, it
 allows better interoperability and has a connection string in more "standard"
 way. Some examples:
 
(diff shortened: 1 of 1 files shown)


WisdomPill (changes requested):
Thanks for pointing it out, now that you are at it can you add an example with `"OPTIONS"`?

WisdomPill (approved):

GrazingScientist:
Thanks for kicking me! :)

I left the previous code block with the `OPTIONS` example in the documentation, because it shows that the `OPTIONS` password can also be used without the ACL credentials.

WisdomPill (approved):

PR #484, opened 2020-12-09
Use tox-gh-actions to reduce config effort.

jezdez:
This is related to the current effort to port all Jazzband projects to GitHub Actions and to use the same best practices to setup GHA as much as possible.


hugovk (approved):
Looks good!

PR #490, opened 2021-01-11
cache delete returns a boolean since django version 3.1 closes #486

### LINKED ISSUE #486: cache.delete returns an integer instead of a boolean
Hello, thanks for providing this great package.

I've just seen in the Django documentation (Version 3.1.) that the `cache.delete` API returns a boolean value [since version 3.1](https://docs.djangoproject.com/en/3.1/topics/cache/#django.core.caches.cache.delete):

Screenshot 2020-12-11 at 08 51 41

When I delete a cache key using the django-redis package it returns an integer instead of a boolean. Is this something that should be changed?

**Example:**
Using one of the built-in cache backends from Django:
```python
cache.delete("foobar")
>>> True
```

Using the django-redis package:
```python
cache.delete("foobar")
>>> 1
```

pinoatrome:
Hi, many thanks for this package.
provided my first (little) contribution from this issue marked as 'good first issue'.
HTH
GG

### DIFF
--- a/django_redis/cache.py
+++ b/django_redis/cache.py
@@ -1,6 +1,7 @@
 import functools
 import logging
 
+from django import VERSION as DJANGO_VERSION
 from django.conf import settings
 from django.core.cache.backends.base import BaseCache
 from django.utils.module_loading import import_string
@@ -95,7 +96,9 @@ class RedisCache(BaseCache):
 
     @omit_exception
     def delete(self, *args, **kwargs):
-        return self.client.delete(*args, **kwargs)
+        """returns a boolean instead of int since django version 3.1"""
+        result = self.client.delete(*args, **kwargs)
+        return bool(result) if DJANGO_VERSION >= (3, 1, 0) else result
 
     @omit_exception
     def delete_pattern(self, *args, **kwargs):
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -355,6 +355,28 @@ class DjangoRedisCacheTests(unittest.TestCase):
         res = self.cache.delete("a")
         self.assertFalse(bool(res))
 
+    @patch("django_redis.cache.DJANGO_VERSION", (3, 1, 0, "final", 0))
+    def test_delete_return_value_type_new31(self):
+        """delete() returns a boolean instead of int since django version 3.1"""
+        self.cache.set("a", 1)
+        res = self.cache.delete("a")
+        self.assertEqual(type(res), bool)
+        self.assertTrue(res)
+        res = self.cache.delete("b")
+        self.assertEqual(type(res), bool)
+        self.assertFalse(res)
+
+    @patch("django_redis.cache.DJANGO_VERSION", (3, 0, 1, "final", 0))
+    def test_delete_return_value_type_before31(self):
+        """delete() returns a int before django version 3.1"""
+        self.cache.set("a", 1)
+        res = self.cache.delete("a")
+        self.assertEqual(type(res), int)
+        self.assertEqual(res, 1)
+        res = self.cache.delete("b")
+        self.assertEqual(type(res), int)
+        self.assertEqual(res, 0)
+
     def test_delete_many(self):
         self.cache.set_many({"a": 1, "b": 2, "c": 3})
         res = self.cache.delete_many(["a", "b"])


WisdomPill:
By adding closes message at the end it will close the issue as soon as this pr is merged

tests/test_backend.py
```diff
@@ -355,6 +355,23 @@ def test_delete(self):
         res = self.cache.delete("a")
         self.assertFalse(bool(res))
 
+    def test_delete_return_value(self):
+        """returns a boolean instead of int since django version 3.1"""
+        self.cache._dj_ver_maj = 4
+        self.cache._dj_ver_min = 0
+        self.cache.set("a", 1)
+        res = self.cache.delete("a")
+        self.assertTrue(res is True)
```
WisdomPill, line 364: there is no need for `assertTrue` with `res is True`, just `assertTrue(res)` is enough

tests/test_backend.py
```diff
@@ -355,6 +355,23 @@ def test_delete(self):
...
+        res = self.cache.delete("b")
+        self.assertTrue(res is False)
```
WisdomPill, line 366: Same here with `assertFalse`

tests/test_backend.py
```diff
@@ -355,6 +355,23 @@ def test_delete(self):
...
+        self.cache._dj_ver_maj = 3
+        self.cache._dj_ver_min = 0
+        self.cache.set("a", 1)
+        res = self.cache.delete("a")
+        self.assertTrue(res == 1)
```
WisdomPill, line 371: Here you can use `assertEqual`

django_redis/cache.py
```diff
@@ -95,7 +98,14 @@ def _get(self, key, default, version, client):
 
     @omit_exception
     def delete(self, *args, **kwargs):
-        return self.client.delete(*args, **kwargs)
+        """returns a boolean instead of int since django version 3.1"""
+        result = self.client.delete(*args, **kwargs)
+        return (
```
WisdomPill, line 103: You can actually compare tuples so `DJANGO_VERSION` > (3, 1, 0) should work

WisdomPill (changes requested):
Thanks for the help, there are a few things to fix but it is should be fairly simple to fix them.

tests/test_backend.py
pinoatrome: done - added a check on the type
```
        self.assertEqual(type(res), bool)
        self.assertTrue(res)
```

tests/test_backend.py
pinoatrome: done - added a check on the type
```
        self.assertEqual(type(res), bool)
        self.assertFalse(res)
```

django_redis/cache.py
pinoatrome: done
`        return bool(result) if self._dj_ver >= (3, 1, 0) else result`

tests/test_backend.py
pinoatrome: done - added a check on the type
```
        self.assertEqual(type(res), int)
        self.assertEqual(res, 1)
```

pinoatrome:
done, amended the code as requested and squashed the commits. 

I've noticed the 'closes' message on the PR, thanks for your availability and patience.

what's next?
sentinel supports is marked with 'needs help'... maybe I can help?

Plus I'd like to share a client implementation that emulates a RedisCache client ... without connecting to a Redis BE instance actually, it wraps a LocMem backend instead.
It is useful for django unit tests scenarios because it exposes the same behaviour (for basic operations  like add, set, get, delete, etc...) as a RedisCache backend while using a LocMem cache under the hood.
How to provide/share this piece of code?
GG

django_redis/cache.py
```diff
@@ -61,6 +62,7 @@ def __init__(self, server, params):
             if self._log_ignored_exceptions
             else None
         )
+        self._dj_ver = DJANGO_VERSION[:3]
```
WisdomPill, line 65: No need to store this value anywhere, it can be mocked in tests

WisdomPill (changes requested):
A small change is required, nice that you have added documentation by the way.

P.S. Sorry that it took long for this review, I forgot about the notification

django_redis/cache.py
pinoatrome: done.

```
   @patch("django_redis.cache.DJANGO_VERSION", (3, 0, 1, 'final', 0))
    def test_delete_return_value_type_before31(self):
```

WisdomPill (approved):
Thanks for the contribution!

pinoatrome:
> By adding closes message at the end it will close the issue as soon as this pr is merged

Hi, thanks for accepting this little contribution. 
A(another) newbie question: I noticed that even if the PR is merged the #486 is still open, shouldn't it be closed automatically?

PR #492, opened 2021-01-12
refactor: ensure close method is thread safe closes #491

### LINKED ISSUE #491: cache.close() modifies protected attribute of ConnectionPool
Hi. Thanks for this package.
I don't understant why does the method close of the DefaultClient access and modify a protected attribute of ConnectionPool in order to close the connections?
https://github.com/jazzband/django-redis/blob/e48d08b448a311339fa1fbb8f05468863b0727cc/django_redis/client/default.py#L571
Why not simply call the method disconnect of ConnectionPool?
This manipulation of a private attribute is not thread safe.

AnesFoufa:
call the method disconnect of ConnectionPool instead of modifying protected attribute

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -567,10 +567,8 @@ class DefaultClient:
 
     def close(self, **kwargs):
         if getattr(settings, "DJANGO_REDIS_CLOSE_CONNECTION", False):
-            for i in range(len(self._clients)):
-                for c in self._clients[i].connection_pool._available_connections:
-                    c.disconnect()
-                self._clients[i] = None
+            for client in self._clients:
+                client.connection_pool.disconnect()
 
     def touch(self, key, timeout=DEFAULT_TIMEOUT, version=None, client=None):
         """
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -569,7 +569,8 @@ class DefaultClient:
         if getattr(settings, "DJANGO_REDIS_CLOSE_CONNECTION", False):
             for i in range(len(self._clients)):
                 client = self._clients[i]
-                self._disconnect_client(client)
+                if client is not None:
+                    client.connection_pool.disconnect()
                 self._clients[i] = None
 
     def touch(self, key, timeout=DEFAULT_TIMEOUT, version=None, client=None):
@@ -590,9 +591,3 @@ class DefaultClient:
             # Convert to milliseconds
             timeout = int(timeout * 1000)
             return bool(client.pexpire(key, timeout))
-
-    @staticmethod
-    def _disconnect_client(client):
-        if client is not None:
-            for connection in client.connection_pool._available_connections:
-                connection.disconnect()
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -572,7 +572,7 @@ class DefaultClient:
                 if client is not None:
                     for connection in client.connection_pool._available_connections:
                         connection.disconnect()
-                self._clients[i] is None
+                self._clients[i] = None
 
     def touch(self, key, timeout=DEFAULT_TIMEOUT, version=None, client=None):
         """
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -568,9 +568,11 @@ class DefaultClient:
     def close(self, **kwargs):
         if getattr(settings, "DJANGO_REDIS_CLOSE_CONNECTION", False):
             for i in range(len(self._clients)):
-                for c in self._clients[i].connection_pool._available_connections:
-                    c.disconnect()
-                self._clients[i] = None
+                client = self._clients[i]
+                if client is not None:
+                    for connection in client.connection_pool._available_connections:
+                        connection.disconnect()
+                self._clients[i] is None
 
     def touch(self, key, timeout=DEFAULT_TIMEOUT, version=None, client=None):
         """
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -568,9 +568,8 @@ class DefaultClient:
     def close(self, **kwargs):
         if getattr(settings, "DJANGO_REDIS_CLOSE_CONNECTION", False):
             for i in range(len(self._clients)):
-                client = self._clients[i]
-                if client is not None:
-                    client.connection_pool.disconnect()
+                for c in self._clients[i].connection_pool._available_connections:
+                    c.disconnect()
                 self._clients[i] = None
 
     def touch(self, key, timeout=DEFAULT_TIMEOUT, version=None, client=None):
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -569,9 +569,7 @@ class DefaultClient:
         if getattr(settings, "DJANGO_REDIS_CLOSE_CONNECTION", False):
             for i in range(len(self._clients)):
                 client = self._clients[i]
-                if client is not None:
-                    for connection in client.connection_pool._available_connections:
-                        connection.disconnect()
+                self._disconnect_client(client)
                 self._clients[i] = None
 
     def touch(self, key, timeout=DEFAULT_TIMEOUT, version=None, client=None):
@@ -592,3 +590,9 @@ class DefaultClient:
             # Convert to milliseconds
             timeout = int(timeout * 1000)
             return bool(client.pexpire(key, timeout))
+
+    @staticmethod
+    def _disconnect_client(client):
+        if client is not None:
+            for connection in client.connection_pool._available_connections:
+                connection.disconnect()
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -703,11 +703,9 @@ class DjangoRedisCacheTests(unittest.TestCase):
 
 
 class DjangoRedisCacheTestsWithClose(DjangoRedisCacheTests):
+    @override_settings(DJANGO_REDIS_CLOSE_CONNECTION=True)
     def test_close(self):
-        with override_settings(
-            CACHE=settings.CACHES, DJANGO_REDIS_CLOSE_CONNECTION=True
-        ):
-            super().test_close()
+        super().test_close()
 
 
 class DjangoRedisCacheWithCloseAndBlockingConnectionPool(
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -559,6 +559,7 @@ class DjangoRedisCacheTests(unittest.TestCase):
             "*foo-a*", itersize=expected_count
         )
 
+    @override_settings(DJANGO_REDIS_CLOSE_CONNECTION=True)
     def test_close(self):
         self.cache.set("f", "1")
         self.cache.close()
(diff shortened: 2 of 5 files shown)


AnesFoufa:
Issus https://github.com/jazzband/django-redis/issues/491

WisdomPill:
Hello @AnesFoufa!
Thanks for the contribution.

Just a couple of questions.

Are you sure it does what you mean? Can you add a test for it?

AnesFoufa:
@WisdomPill : I didn't add tests because it is a refactor.
I added tests to test the modified method, which was not covered by the existing tests, and I discovered an other bug on the same method that I fixed.

AnesFoufa:
The failing test doesn't seem deterministic because it passes ofttimes. https://github.com/AnesFoufa/django-redis/runs/1694948805

tests/test_backend.py
```diff
@@ -288,7 +288,7 @@ def test_set_add(self):
 
         res = cache.get("add_key")
         self.assertEqual(res, "Initial value")
-
+        cache.delete("other_key")
```
WisdomPill, line 291: Why you added this?

tests/test_backend.py
```diff
@@ -702,6 +702,32 @@ def test_touch_default_timeout(self):
         self.assertEqual(self.cache.get("test_key"), "foo")
 
 
+class DjangoRedisCacheTestsWithClose(DjangoRedisCacheTests):
```
WisdomPill, line 705: This can be a decorator of the class

tests/test_sqlite.py
```diff
@@ -1,10 +1,12 @@
 SECRET_KEY = "django_tests_secret_key"
-
 CACHES = {
     "default": {
         "BACKEND": "django_redis.cache.RedisCache",
         "LOCATION": ["redis://127.0.0.1:6379?db=1", "redis://127.0.0.1:6379?db=1"],
-        "OPTIONS": {"CLIENT_CLASS": "django_redis.client.DefaultClient"},
+        "OPTIONS": {
```
WisdomPill, line 6: Why are you modifying the settings for all the tests?

tests/test_sqlite.py
```diff
@@ -25,3 +27,4 @@
 }
 
 INSTALLED_APPS = ["django.contrib.sessions"]
+DJANGO_REDIS_CLOSE_CONNECTION = False
```
WisdomPill, line 30: I think this should not be added, even if it does not do any harm, otherwise all the default values has to be added in all the tests, which I think it is not worth it

WisdomPill (changes requested):
Have a look at how you should change the settings for tests [here](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#overriding-settings)

Apart from that, if the connection closing was not tested it needs to be tested otherwise we can't include a modification/refactoring because it can't be proven that the behavior is still the same

tests/test_sqlite.py
AnesFoufa: This is a mistake, i restore

tests/test_sqlite.py
AnesFoufa: I fix.

AnesFoufa:
For modifying the settings in tests, I use override_settings, should I do it differently?
As for close method, I added a test before making any change and the method raised an exception. I fixed this bug.

tests/test_backend.py
AnesFoufa: I don't know how, could you share some documentation?

tests/test_backend.py
AnesFoufa: In case the clear cache in the setUp fails.

tests/test_sqlite.py
WisdomPill: I already shared [here](https://github.com/jazzband/django-redis/pull/492#pullrequestreview-567203974)

tests/test_backend.py
WisdomPill: I see, I think that `setUp` is a very bad thing, exceptions should never pass silently, especially on tests, I suggest adding the `self.cache.clear()` on `tearDown()` method and remove the try catch block, to see if there are any exceptions and to save the whole suite of tests

WisdomPill:
I see that `test_close` does not have any assertion, maybe they thought it would throw some exceptions, I think the `client._sock` property could be checked since it is set to `None` on the disconnect call [here](https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/connection.py#L663)

WisdomPill:
> For modifying the settings in tests, I use override_settings, should I do it differently?
> As for close method, I added a test before making any change and the method raised an exception. I fixed this bug.

I was saying that it can be done in a more elegant way just adding a decorator instead of the context manager.

tests/test_sqlite.py
AnesFoufa: If you mean `with self.settings(LOGIN_URL='/other/login/'):` it only works with django.tests.TestCase, not unittest.TestCase and django.tests.TestCase requires a database.

WisdomPill: You can use, `SimpleTestCase` (have a look [here](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#simpletestcase)) like it is done in `django` (have a look [here](class DummyCacheTests(SimpleTestCase):))

AnesFoufa:
> I see that `test_close` does not have any assertion, maybe they thought it would throw some exceptions, I think the `client._sock` property could be checked since it is set to `None` on the disconnect call [here](https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/connection.py#L663)

The disconnect method is supposed to run successfully even if client._sock is None

tests/test_sqlite.py
AnesFoufa: > You can use, `SimpleTestCase` (have a look [here](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#simpletestcase)) like it is done in `django` (have a look [here](https://github.com/django/django/blob/83fcfc9ec8610540948815e127101f1206562ead/tests/cache/tests.py#L82))

My classes would be the only ones that extend SimpleTestCase. I think my test classes should stick to unittest.TestCase like the other ones or all the test classes should extend djange.test.SimpleTestCase

AnesFoufa:
> I see that `test_close` does not have any assertion, maybe they thought it would throw some exceptions, I think the `client._sock` property could be checked since it is set to `None` on the disconnect call [here](https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/connection.py#L663)

The legacy test_close did throw an exception with the setting DJANGO_REDIS_CLOSE_CONNECTION at True.

WisdomPill:
> > I see that `test_close` does not have any assertion, maybe they thought it would throw some exceptions, I think the `client._sock` property could be checked since it is set to `None` on the disconnect call [here](https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/connection.py#L663)
> 
> The legacy test_close did throw an exception with the setting DJANGO_REDIS_CLOSE_CONNECTION at True.

Can you post it?

WisdomPill:
> > I see that `test_close` does not have any assertion, maybe they thought it would throw some exceptions, I think the `client._sock` property could be checked since it is set to `None` on the disconnect call [here](https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/connection.py#L663)
> 
> The disconnect method is supposed to run successfully even if client._sock is None

I understand I said that it could be asserted that `connection._sock is None`

tests/test_sqlite.py
WisdomPill: You can do with plain `unittest.TestCase` or switch everything to `django.test.SimpleTestCase`, I have a plan to do some refactoring on the tests anyway

tests/test_backend.py
WisdomPill: Also this one could be skipped, I have a plan to refactor the test suite

tests/test_sqlite.py
AnesFoufa: I agree, the tests should be refactored.

AnesFoufa:
> > > I see that `test_close` does not have any assertion, maybe they thought it would throw some exceptions, I think the `client._sock` property could be checked since it is set to `None` on the disconnect call [here](https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/connection.py#L663)
> > 
> > 
> > The legacy test_close did throw an exception with the setting DJANGO_REDIS_CLOSE_CONNECTION at True.
> 
> Can you post it?

I just did and you can see the failures here https://github.com/jazzband/django-redis/actions/runs/483232890

WisdomPill:
I see, so that the client is not checked because it could be `None` also the control of available_connection is the sameish as [here](https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/connection.py#L1221)

I think that just adding your tests it is okay, then I will do some refactoring

AnesFoufa:
> I see, so that the client is not checked because it could be `None` also the control of available_connection is the sameish as [here](https://github.com/andymccurdy/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/connection.py#L1221)
> 
> I think that just adding your tests it is okay, then I will do some refactoring

So I only added my the test that reveals the client at None bug in close and my  proposed (minimal) solution. If it is merged, I'll add a separate pull request to deal with the BlockingCOnnectionPool bug.

tests/test_backend.py
WisdomPill: Remove this, I will fix it in the refactor, if this has to stay then a lot of other cleanup has to be done

tests/test_backend.py
WisdomPill: Do not subclass `DjangoRedisCacheTests`, it will run tests twice, one for the parent class and one for the child

django_redis/client/default.py
```diff
@@ -568,8 +568,8 @@ def make_pattern(self, pattern, version=None, prefix=None):
     def close(self, **kwargs):
         if getattr(settings, "DJANGO_REDIS_CLOSE_CONNECTION", False):
             for i in range(len(self._clients)):
-                for c in self._clients[i].connection_pool._available_connections:
-                    c.disconnect()
+                client = self._clients[i]
+                self._disconnect_client(client)
```
WisdomPill, line 572: I do not think a separate method is needed, just check if it is `None` and call `.disconnect()` from the connection_pool

WisdomPill (changes requested):

WisdomPill:
> So I only added my the test that reveals the client at None bug in close and my proposed (minimal) solution. If it is merged, I'll add a separate pull request to deal with the BlockingCOnnectionPool bug.

For the blocking connection I would add the test and skip it and leave a comment on top for the explanation

AnesFoufa:
> > So I only added my the test that reveals the client at None bug in close and my proposed (minimal) solution. If it is merged, I'll add a separate pull request to deal with the BlockingConnectionPool bug.
> 
> For the blocking connection I would add the test and skip it and leave a comment on top for the explanation

This test needs to run in a separate process so I need an other settings file, an other test command and an other job in nox. It's pretty heavy so I prefer an other pull request. This one already fixes a problem.

django_redis/client/default.py
AnesFoufa: OK I do

AnesFoufa: As for cache.delete("other_key"), it is needed to pass these tests. I solve this problem in this PR https://github.com/jazzband/django-redis/pull/494

WisdomPill (approved):
Thanks, I think it is ready now

AnesFoufa:
Thank you for taking your time to review my first contribution.

PR #493, opened 2021-01-13
use redis service instead of 3rd party supercharge/redis-github-action@1.1.0

terencehonles:
Pulling this change out of #460.

I'm not sure why the action was used, but it's simpler and better to use the service so that the health check is performed (even though it's unlikely it wouldn't be healthy by the time tox runs)


WisdomPill (approved):
Thanks for contributing!

PR #494, opened 2021-01-14
Fix clear shard client

AnesFoufa:
This pull request tests the method clear fir all the clients. This method raises an NotImplementedError for ShardClient as seen here
https://github.com/AnesFoufa/django-redis/actions/runs/484855381
So I added an basic implementation that passes the test.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -55,8 +55,10 @@ class DjangoRedisCacheTestEscapePrefix(unittest.TestCase):
         self.addCleanup(cm.disable)
 
         self.cache = caches["default"]
-        self.cache.clear()
         self.other = caches["with_prefix"]
+
+    def tearDown(self):
+        self.cache.clear()
         self.other.clear()
 
     def test_delete_pattern(self):
@@ -92,6 +94,8 @@ class DjangoRedisCacheTestCustomKeyFunction(unittest.TestCase):
         self.addCleanup(cm.disable)
 
         self.cache = caches["default"]
+
+    def tearDown(self):
         self.cache.clear()
 
     def test_custom_key_function(self):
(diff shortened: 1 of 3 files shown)


tests/test_backend.py
```diff
@@ -1218,3 +1205,20 @@ def test_delete_pattern_calls_delete_for_given_keys(
         client.delete_pattern(pattern="foo*")
 
         connection.delete.assert_called_once_with(*connection.scan_iter.return_value)
+
+
+class TestClearClient(unittest.TestCase):
+    cache_name = "default"
```
WisdomPill, line 1211: I see no need on putting class variables that are used only once

tests/test_backend.py
```diff
@@ -55,15 +55,9 @@ def setUp(self):
         self.addCleanup(cm.disable)
 
         self.cache = caches["default"]
-        try:
-            self.cache.clear()
-        except Exception:
-            pass
+        self.cache.clear()
```
WisdomPill, line 58: please put it on `tearDown` it is easier to understand and complies with the common practice

WisdomPill (changes requested):

tests/test_backend.py
AnesFoufa: Done.

tests/test_backend.py
AnesFoufa: I can do it, but if we need to run the same with different cache, a different key or a different value, it can be easily done by extending this class and overriding the class attributes. Than enables me to easily reuse this test here: https://github.com/jazzband/django-redis/blob/e109e4e3ce655ae4d4217150997915a43b941a55/tests/test_backend.py#L1233

WisdomPill: It is not readable the same way, if you subclass a test class it is not straightforward to understand that the test is ran twice with different parameters, readability is more important than writing less code. Same goes with class variables, a test should be self explanatory without knowing all the customizations of the class. These two classes I think can go in `DjangoRedisCacheTests` with two different methods

WisdomPill (changes requested):

WisdomPill:
Thanks for the contribution!

WisdomPill (approved):

PR #508, opened 2021-04-03
[Update] expire and ttl are replaced by pexipre and pttl to provide milli-seconds precision

### LINKED ISSUE #333: adopt pexpire, pttl to align accuracy of set operation
To whom might concern this case,

I found the `set` API adopting ms in specifying expiration of redis `set()`. Therefore, the caller can use millisecond to expire the cache, just like
```
cache.set('key', 'value', timeout=1.5)
```

But, the `expire` and `ttl` API did not align the accuracy. Will we consider aligning the accuracy?

Thank you.

### LINKED ISSUE #505: touch method converting seconds to milliseconds and then using pexpire instead of expire.
https://github.com/jazzband/django-redis/blob/afb00223729d5caafa8b63f3684532aaa63f1963/django_redis/client/default.py#L604-L606

My suggestions : 
```
  def touch(self, key, timeout=DEFAULT_TIMEOUT, version=None, client=None):
        ...
        if timeout is None:
            return bool(client.persist(key))
        else:
            return bool(client.expire(key, timeout))
```

abbasidaniyal:
Fixes #333 
Fixes #505

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -350,9 +350,11 @@ It returns:
     >>> from django.core.cache import cache
     >>> cache.set("foo", "value", timeout=25)
     >>> cache.ttl("foo")
-    25
+    25.0
     >>> cache.ttl("not-existent")
-    0
+    0.0
+
+Note: ttl previously returned an int. Now it returns a float.
 
 Expire & Persist
 ~~~~~~~~~~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -350,11 +350,21 @@ It returns:
     >>> from django.core.cache import cache
     >>> cache.set("foo", "value", timeout=25)
     >>> cache.ttl("foo")
-    25.0
+    25
     >>> cache.ttl("not-existent")
-    0.0
+    0
 
-Note: ttl previously returned an int. Now it returns a float.
+With Redis, you can access to ttl of any stored key in milliseconds, for it, django-redis
+exposes ``pttl`` function.
+
+.. code-block:: pycon
+
+    >>> from django.core.cache import cache
+    >>> cache.set("foo", "value", timeout=25)
+    >>> cache.pttl("foo")
+    25000
+    >>> cache.pttl("not-existent")
+    0
 
 Expire & Persist
 ~~~~~~~~~~~~~~~~
@@ -377,6 +387,14 @@ specify a new expiration timeout using the ``persist`` and ``expire`` methods:
     >>> cache.expire("foo", timeout=5)
     >>> cache.ttl("foo")
     5
+
+The ``pexpire`` method can be used to provide millisecond precision:
+.. code-block:: pycon
+
+    >>> cache.set("foo", "bar", timeout=22)
+    >>> cache.pexpire("foo", timeout=5500)
+    >>> cache.pttl("foo")
+    5500
 
 Locks
 ~~~~~
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -581,7 +581,8 @@ class DefaultClient:
         """
         Executes TTL redis command and return the "time-to-live" of specified key.
         If key is a non volatile key, it returns None.
-        Warning: ttl previously returned an int. Now it returns a float.
+        Note: ttl previously return the number of seconds in an integer. Now it
+        returns the number of seconds with decimals which represent milliseconds in a float.
         """
         if client is None:
             client = self.get_client(write=False)
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -581,8 +581,8 @@ class DefaultClient:
         """
         Executes TTL redis command and return the "time-to-live" of specified key.
         If key is a non volatile key, it returns None.
-        Note: ttl previously return the number of seconds in an integer. Now it
-        returns the number of seconds with decimals which represent milliseconds in a float.
+        Note: ttl previously return the number of seconds in an integer. Now it returns
+        the number of seconds with decimals which represent milliseconds in a float.
         """
         if client is None:
             client = self.get_client(write=False)
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -592,7 +592,7 @@ class DefaultClient:
         t = client.pttl(key)
 
         if t >= 0:
-            return t / 1000
+            return round(t / 1000, 3)
         elif t == -1:
             return None
         elif t == -2:
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -288,7 +288,15 @@ class DefaultClient:
         key = self.make_key(key, version=version)
 
         if client.exists(key):
-            timeout = int(timeout * 1000)
+            client.expire(key, timeout)
+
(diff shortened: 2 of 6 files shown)


django_redis/client/default.py
```diff
@@ -489,10 +490,10 @@ def ttl(self, key, version=None, client=None):
         if not client.exists(key):
             return 0
 
-        t = client.ttl(key)
+        t = client.pttl(key)
```
WisdomPill, line 493: I would put a warning and highlight in README that `ttl` now returns a float. I would also round it to 3 decimals

tests/test_backend.py
```diff
@@ -591,9 +591,9 @@ def test_ttl(self):
         ttl = cache.ttl("foo")
 
         if isinstance(cache.client, herd.HerdClient):
-            self.assertAlmostEqual(ttl, 12)
+            self.assertAlmostEqual(ttl, 12, delta=0.001)
```
WisdomPill, line 594: This should be removed in favor of rounding the number of seconds returning in the function.

WisdomPill (changes requested):

WisdomPill:
Also tests for non integer timeout should be added

tests/test_backend.py
abbasidaniyal: After rounding the response from pttl and removing delta, I'm getting 
```
AssertionError: 9.999 != 10 within 7 places (0.0009999999999994458 difference)
```
Since the value of places defaults to 7. This seems to be a floating point precision problem, thus supplying delta seems necessary here.

abbasidaniyal: However, that leads to errors when comparing values like,
```AssertionError: 5.499 != 5.5 within 0.001 delta (0.001000000000000334 difference)```

Comment (abbasidaniyal) [reply tests/test_backend.py:594]: This is the logic with rounding the value returned from redis
```
t = client.pttl(key)
...
round(t / 1000, 3)
```

Comment (WisdomPill) [reply tests/test_backend.py:594]: I think delta should remain and a larger delta is needed, of course the code is right but we cannot control the delay between the check and redis response at the millisecond precision because also everybody does not have the biggest resources to get close to it.

File: django_redis/client/default.py (hunk 1 above)
Comment (abbasidaniyal) [reply django_redis/client/default.py:493]: How should a warning be displaed? Using the warnings.warn function or something else?

Comment (WisdomPill) [reply django_redis/client/default.py:493]: warning and docs

Comment (WisdomPill) [reply django_redis/client/default.py:493]: Also a comment on the tests would be helpful

File: README.rst [hunk 3]
```diff
@@ -350,9 +350,11 @@ It returns:
     >>> from django.core.cache import cache
     >>> cache.set("foo", "value", timeout=25)
     >>> cache.ttl("foo")
-    25
+    25.0
     >>> cache.ttl("not-existent")
-    0
+    0.0
+
+Note: ttl previously returned an int. Now it returns a float.
```
WisdomPill, line 357: Note: til previously return the number of seconds in an integer. Now it returns the number of seconds with decimals which represent milliseconds in a float.

django_redis/client/default.py
```diff
@@ -482,6 +482,7 @@ def ttl(self, key, version=None, client=None):
         """
         Executes TTL redis command and return the "time-to-live" of specified key.
         If key is a non volatile key, it returns None.
+        Warning: ttl previously returned an int. Now it returns a float.
```
WisdomPill, line 485: Note: til previously return the number of seconds in an integer. Now it returns the number of seconds with decimals which represent milliseconds in a float.

WisdomPill (changes requested):
I thought you would use `warning` package, have a look [here](https://docs.python.org/3/library/warnings.html). Does it work with doctoring as well?
I never used warnings before, so I might be wrong

abbasidaniyal:
> I thought you would use `warning` package, have a look [here](https://docs.python.org/3/library/warnings.html). Does it work with doctoring as well?
> I never used warnings before, so I might be wrong

I was but I couldn't find any other useage of the warning package. Furthermore, the only warning out of [these](https://docs.python.org/3/library/warnings.html#warning-categories) that would make some sense would be `FutureWarning`. The warning should be added first and the functionality should be release in a future version. Still don't think it does much good to our case.

WisdomPill:
I think `UserWarning` or `Warning` might work.
After all it is kind of a user code generated warning.

Regardless of which warning, a warning is needed to make the user aware of the change

WisdomPill (changes requested):
I think UserWarning or Warning might work.
After all it is kind of a user code generated warning.

Regardless of which warning, a warning is needed to make the user aware of the change

WisdomPill:
@terencehonles do you mind giving your thoughts about this PR?

I would love to have this included in the next release as well

terencehonles:
> @terencehonles do you mind giving your thoughts about this PR?
> 
> I would love to have this included in the next release as well

I took a glance at it, it's probably fine as is, but just making sure there's a top level comment to make it obvious I replied in an existing comment https://github.com/jazzband/django-redis/pull/508#discussion_r625244588

I have a slight preference to add the `pttl` method, but it's not that big of a deal if you like or prefer it this way.

django_redis/client/default.py
WisdomPill: I took some time to think about it and I agree with @terencehonles, it would be better to add `pexpire` and `pttl`

WisdomPill: @abbasidaniyal what do you think? are you able to work on this quickly? I would love to add it to the next release. For the moment I will write the `CHANGELOG` and wait a day or two for your response, otherwise we will release without this change.

abbasidaniyal: Yes sure. Working on it

abbasidaniyal: @WisdomPill Let me know if I'm on the right path. If not, do let me know what changes I should make!

WisdomPill (changes requested):
Seems like tests are not passing, will create the `CHANGELOG` anyway so that we can release asap, this will probably go to the next release

abbasidaniyal:
> Seems like tests are not passing, will create the `CHANGELOG` anyway so that we can release asap, this will probably go to the next release

The tests are failing due to precision
```
AssertionError: 5499 != 5500 within 7 places
```

I am thinking of adding the `delta` param to the `assertAlmostEqual` method. However, it does not feel like a clean approach. Still, doing it with `delta=1` for now. Let me know what you feel.

WisdomPill:
Well we are talking milliseconds so maybe just add a comment before to explain why the delta is needed and it will be good to be approved

WisdomPill (changes requested):
I guess the delta should be increased to 10 in the millisecond calls, `pexpire` and `pttl`, so that the test would pass all the time

WisdomPill (changes requested):
LGTM! Please rebase and edit `CHANGES` file so that this would be included in the next release

WisdomPill (approved):
After tests pass I will merge it, thanks for the contribution!

PR #513, opened 2021-04-07
Split linters

WisdomPill:
- split linters
- fixed cache hash
- added cache for lint job


WisdomPill:
Mybae listing should be done with installing python 3.6 since it is the lowest version possible, so we could ensure that at least with that version it is compatible?

terencehonles:
> Mybae listing should be done with installing python 3.6 since it is the lowest version possible, so we could ensure that at least with that version it is compatible?

I'm not sure if any of the lint tools actually need to be run on an older version in order to check things for older tools (I believe they usually have flags for what to lint for)

terencehonles:
I don't think we need to have that specific of a cache key. I made my alternative suggestion in #519 (thanks for catching it)

WisdomPill:
Let's discuss before opening PRs, by the way I think we should enable the discuss tab on this repo.
And also let's write everything on one PR to keep the discussion in one place.

The lint cache keys are separate because the lint packages are installed separately, so having a unique cache key would always result in the lint package that is needed to be dowloaded because it is not found in cache. I do not think it is a problem to have 4 different cache keys, cache limit is way higher.

terencehonles:
I created a separate PR because I think it can be merged in while we continue to discuss this. I think there are potentially bigger things to discuss with this PR and the CI configuration, and I don't know why we'd want to create separate cache keys so I'm pretty sure #519 can be merged in as is. Do you have any objections to it? I marked you as a co-author since you pointed out the issue and fixed it.

WisdomPill:
I don't think that PR should be merged. I think we should keep different cache keys otherwise lint packages will not be cached. As a matter of fact they might be cached randomly depending on which job finishes last

terencehonles:
> I don't think that PR should be merged. I think we should keep different cache keys otherwise lint packages will not be cached. As a matter of fact they might be cached randomly depending on which job finishes last

I may misunderstand how that works, but I may be following your logic. Did you have documentation of that behavior?

WisdomPill:
Well, it is not documentation but experience from other repos, in this case there is no proof because when tox installs dependencies it does not show pip's output.

If you have a look at Install dependencies step in some runs of #519 you will see a mix of `Downloading` and `Using cached`, like the following.

```
2021-04-08T06:55:08.3660021Z Collecting tox
2021-04-08T06:55:08.3692850Z   Using cached tox-3.23.0-py2.py3-none-any.whl (85 kB)
2021-04-08T06:55:08.4526402Z Collecting tox-gh-actions
2021-04-08T06:55:08.4902160Z   Downloading tox_gh_actions-2.5.0-py2.py3-none-any.whl (8.0 kB)
2021-04-08T06:55:08.5804299Z Collecting pluggy>=0.12.0
2021-04-08T06:55:08.5833981Z   Using cached pluggy-0.13.1-py2.py3-none-any.whl (18 kB)
2021-04-08T06:55:08.6714264Z Collecting py>=1.4.17
2021-04-08T06:55:08.6746787Z   Using cached py-1.10.0-py2.py3-none-any.whl (97 kB)
```

In this case it is happening because the cache key is not present, it is just an example to show what happens to packages not present in github cache or system cache. 
But the same applies in the case of having the same key, `black`, `flake8` and `isort` with their dependencies might not be cached, hence they will be downloaded.

terencehonles:
Based on that we should also be caching by the matrix parameters DJANGO/REDIS too... not just `tox`

WisdomPill:
Well, you are right, I see you got the point 😂 . Let me fix it right now

WisdomPill:
With -v flag we could see pip installation output, maybe it is a good idea to turn it on?

And why python 3.9 is tested only on latest versions of django and redis?

terencehonles:
> With -v flag we could see pip installation output, maybe it is a good idea to turn it on?
I had removed it because I thought it added too much chatter to tox, and I don't think we normally care about the pip install output

> And why python 3.9 is tested only on latest versions of django and redis?
They are not, everything tests the latest django and redis. 3.9 only tests the *unreleased* django/redis. For the others it's very likely fine to break if/when there is a release rather than testing against unreleased code. Most likely something will break on any of the Pythons rather than a specific version and I don't think we should be pessimistically burning through CI credits.

WisdomPill:
Understood, on that matter, I think master of redis and django should be allowed to fail, is there anything like "allow to fail".

I did not search enough for this feature, but it might not have been released yet.

terencehonles:
There is https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error but the checks are not enforced right now so we can leave it failing for now. I would suggest if something breaks on Django / Redis master we should probably evaluate if we need to fix it (an early heads up) or if it's something that needs to be raised in one of their repos. I don't know how GitHub renders a failed check in that case though as I have never used the feature.

terencehonles:
Regarding the PR in general, please give me some time to pull down the PR and test how it is with tox. This was not _exactly_ what I was thinking and I wanted to see if it was better or worse for local dev using tox. I don't want to degrade local testing in order to have a different CI experience.

WisdomPill:
Well, imho it is not really degrading, local testing stays the same, as a matter of fact I think I have already mentioned that we can keep `testenv:lint` to run all of them locally if needed.

terencehonles:
> Well, imho it is not really degrading, local testing stays the same, as a matter of fact I think I have already mentioned that we can keep `testenv:lint` to run all of them locally if needed.

How were you running it? I cleaned up the `lint` env which was actually running the tests since there was no lint config anymore, and I adjusted it so all the lint tools would share the same virtualenv. We could remove that sharing if you don't think it's needed, but it will at least speed up the workflow locally a little and if the tools share dependencies it will reduce duplication (which is not that big of a deal).

On the CI workflow lint run there was no matix.python-version so I removed that and I added a couple fall back keys for the test cache step, and I fixed the indentation like I had in #519

WisdomPill (commented):
I understand the changes, for me it is okay as long as the lint checks are separate

WisdomPill:
I forgot to save the comment...

I asked why in lint job the version installed is python 3.x and not 3.8 because it is the one used in tox.

I don't remember if I was launching the steps together or not in one command. I probably have not even used tox during testing

terencehonles (approved):

WisdomPill:
I think #519 could be closed and this one merged

terencehonles:
> I forgot to save the comment...
> 
> I asked why in lint job the version installed is python 3.x and not 3.8 because it is the one used in tox.
> 
> I don't remember if I was launching the steps together or not in one command. I probably have not even used tox during testing

I can update it to match. Let me see what it actually runs.

terencehonles:
It looks like it runs on Python 3.9 which is fine. The other config is for tox-gh-actions which we're not using for the lint steps.

WisdomPill:
Maybe it is better to pinpoint the target python because you never know when python 3.10 is coming out and there might be some issues

terencehonles:
Hopefully the tools are testing against pre release Pythons, but we can change it if and when it breaks 🤞

Issue #499, opened 2021-02-10
`incr` bug - deletes key when timeout is None if `client.eval` errors


AceFire6:
I have a Django project set up with my cache config as follows:

```python
CACHES = {
    'default': {
        'BACKEND': 'django_redis_sentinel.cache.RedisSentinelCache',
        'LOCATION': [
            (REDIS_SENTINEL_SERVICE_URL, REDIS_SENTINEL_PORT),
        ],
        'TIMEOUT': REDIS_CACHE_TIMEOUT,
        'OPTIONS': {
            'CLIENT_CLASS': 'django_redis_sentinel.client.SentinelClient',
            'SENTINEL_SERVICE_NAME': REDIS_SENTINEL_SERVICE_NAME,
            'REDIS_CLIENT_KWARGS': {
                'db': REDIS_DATABASE,
                'decode_responses': True,
            },
            'SERIALIZER': 'django_redis.serializers.json.JSONSerializer',
            'MAX_ENTRIES': REDIS_CACHE_MAX_ENTRIES,
        },
    },
}
```

We noticed weird behaviour in a case where we were calling `cache.incr(KEY, float_value)`

The keys were being created with a `TTL` of `0`.

Testing locally, I believe I've found the error. In `DefaultClient._incr`: 
https://github.com/jazzband/django-redis/blob/92896088479cb6710fb410307adb66d27dc9e058/django_redis/client/default.py#L408-L451

Specifically in this section of that code:
https://github.com/jazzband/django-redis/blob/92896088479cb6710fb410307adb66d27dc9e058/django_redis/client/default.py#L433-L447

When the `timeout` value is retrieved `django-redis` uses `client.ttl`. This ignores all the post-processing done in `DefaultClient.ttl`.

`client.ttl` in the case of the `None` timeout (we never want our keys to timeout by default) returns `-1`. In the call to `DefaultClient.set` that results when an error occurs (see above) `timeout` is set to `-1` and as can be seen in the following snippet. The key is deleted when `timeout` is negative and `nx` is not `True`:

https://github.com/jazzband/django-redis/blob/92896088479cb6710fb410307adb66d27dc9e058/django_redis/client/default.py#L139-L153

The simplest solution I've found in my testing is to call `self.ttl` instead of `client.ttl`

WisdomPill:
Hello @AceFire6 thanks for pointing out this issue.

Would you be entered in making a PR to fix this?

annbgn:
[...]

Issue #483, opened 2020-11-14
incr_version and decr_version delete keys with infinite timeout


z0ng33:
When I try to incease or decrease version of some keys with infinite timeout, they just disappear.

After small research, I figured out that problem is in `django_redis.client.default.DefaultClient` method `incr_version`:
```python
try:
    ttl = client.ttl(old_key)
except _main_exceptions as e:
    raise ConnectionInterrupted(connection=client) from e
```

When we get ttl of the current key, redis client returns `-1`. After that, when client trying to set this ttl to the new key, it disappear immediately, because for infinite timeout we need to set `ttl=None`.

I guess, it can be fixed with following:
```python
ttl = client.ttl(old_key)
if ttl == -1:
    ttl = None
```

WisdomPill:
Hello @z0ng33 ,

I did not understand what are you saying?

Do you have any code snippet or unittest to prove what are you saying?
It would be very helpful.

Thanks!

z0ng33:
@WisdomPill It's example from django documentation:
```python
from django.core.cache import cache

cache.set('my_key', 'hello world!')
# redis:
# localhost:6379> keys *
# 1) ":1:my_key"

cache.incr_version('my_key')
# redis:
# localhost:6379> keys *
# (empty array)
#
# expected:
# 1) ":2:my_key"
```

PR #517, opened 2021-04-07
fix incr bug that deletes key

annbgn:
as noted in #499 the bug should be gone after changing client.ttl to self.ttl


WisdomPill:
I would suggest to write a test like it is shown in the issue

WisdomPill:
Also #483 is related

WisdomPill:
Sorry for stepping in but I saw no response for a month almost so I took care of it in #524, there will be new good first issues about typing soon if you'd like to help

PR #519, opened 2021-04-07
fix CI cache key

terencehonles:
Pulling the cache change out of #513 and altering it a little (plus making sure the whitespace matches).

Ignore the whitespace [with this diff](https://github.com/jazzband/django-redis/pull/519/files?w=1)


WisdomPill (commented):
I answered in the original PR

PR #520, opened 2021-04-08
update hashring tests as an example of how to convert to pytest tests

terencehonles:
This PR has been annotated with a brief description how to convert unittest tests to pytest test cases for #521.

### DIFF
--- a/tests/test_hashring.py
+++ b/tests/test_hashring.py
@@ -1,4 +1,4 @@
-import unittest
+import pytest
 
 from django_redis.hash_ring import HashRing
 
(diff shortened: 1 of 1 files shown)


tests/test_hashring.py
```diff
@@ -14,24 +14,21 @@ def __repr__(self):
...
-    def setUp(self):
-        self.node0 = Node(0)
-        self.node1 = Node(1)
-        self.node2 = Node(2)
+@pytest.fixture
+def hash_ring():
+    return HashRing([Node(i) for i in range(3)])
 
-        self.nodes = [self.node0, self.node1, self.node2]
-        self.ring = HashRing(self.nodes)
```
terencehonles, line 24: Unfortunately the diff here is not great, but `setUp()` code can generally be converted to a `pytest.fixture`. They can either return a value or yield a value in order to perform cleanup after a test finishes (to act as a `tearDown()`).

For more information: https://docs.pytest.org/en/stable/fixture.html#quick-example

tests/test_hashring.py
```diff
@@ -14,24 +14,21 @@ def __repr__(self):
...
 
-    def test_hashring(self):
-        ids = []
+def test_hashring(hash_ring):
```
terencehonles, line 22: The `hash_ring` fixture is requested here and references are adjusted in the test function.

tests/test_hashring.py
```diff
@@ -14,24 +14,21 @@ def __repr__(self):
...
-        self.assertEqual(ids, [0, 2, 1, 2, 2, 2, 2, 0, 1, 1])
+    assert ids == [0, 2, 1, 2, 2, 2, 2, 0, 1, 1]
 
-    def test_hashring_brute_force(self):
-        for key in (f"test{x}" for x in range(10000)):
-            self.ring.get_node(key)
+
+def test_hashring_brute_force(hash_ring):
+    for key in (f"test{x}" for x in range(10000)):
+        assert hash_ring.get_node(key)
```
terencehonles, line 34: I'm not sure what the previous test was actually testing other than no exceptions were thrown, but this test now asserts that the value is also truthy.

WisdomPill (commented):
Maybe node could be a `dataclass`? They are available from python 3.6 so it should be okay to use them

tests/test_hashring.py
WisdomPill: That's quite strange, it was not asserting anything, it can be removed as far as I can tell

terencehonles:
> Maybe `Node` could be a `dataclass`? They are available from python 3.6 so it should be okay to use them

Should we leave that conversion out of the PR since that's starting to distract from the point of this PR (an example of converting unittest tests to pytest tests)?

WisdomPill:
Agree, I am merging it then

PR #523, opened 2021-05-01
Fix incr_version of key with no expiration

### LINKED ISSUE #483: incr_version and decr_version delete keys with infinite timeout

WisdomPill:
Fixes version increment of keys with no expiration, that were previously wrongly deleted

fixes: #483

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -679,7 +679,7 @@ class DefaultClient:
             self.do_close_clients()
 
     def do_close_clients(self):
-        """ default implementation: Override in custom client """
+        """default implementation: Override in custom client"""
         num_clients = len(self._clients)
         for idx in range(num_clients):
             self.disconnect(index=idx)
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -206,6 +206,12 @@ class DefaultClient:
 
         try:
             ttl = client.ttl(old_key)
+
+            # redis ttl returns -1 if key does not have a timeout,
+            # so ttl has to be set to None to preserve the absence of a timeout
+            # in the new version as well
+            if ttl == -1:
+                ttl = None
         except _main_exceptions as e:
             raise ConnectionInterrupted(connection=client) from e
 
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -205,13 +205,7 @@ class DefaultClient:
         value = self.get(old_key, version=version, client=client)
 
         try:
-            ttl = client.ttl(old_key)
-
-            # redis ttl returns -1 if key does not have a timeout,
-            # so ttl has to be set to None to preserve the absence of a timeout
-            # in the new version as well
-            if ttl == -1:
-                ttl = None
+            ttl = self.ttl(old_key, version=version, client=client)
         except _main_exceptions as e:
             raise ConnectionInterrupted(connection=client) from e
 
--- a/django_redis/client/sharded.py
+++ b/django_redis/client/sharded.py
@@ -198,6 +198,12 @@ class ShardClient(DefaultClient):
 
         try:
             ttl = client.ttl(old_key)
+
+            # redis ttl returns -1 if key does not have a timeout,
+            # so ttl has to be set to None to preserve the absence of a timeout
+            # in the new version as well
+            if ttl == -1:
+                ttl = None
         except ConnectionError as e:
             raise ConnectionInterrupted(connection=client) from e
 
--- a/django_redis/client/sharded.py
+++ b/django_redis/client/sharded.py
@@ -197,13 +197,7 @@ class ShardClient(DefaultClient):
         value = self.get(old_key, version=version, client=client)
 
         try:
-            ttl = client.ttl(old_key)
-
-            # redis ttl returns -1 if key does not have a timeout,
-            # so ttl has to be set to None to preserve the absence of a timeout
-            # in the new version as well
-            if ttl == -1:
-                ttl = None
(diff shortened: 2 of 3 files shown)


WisdomPill:
No worries it is just mypy coverage that gets fooled by comments

terencehonles (approved):
LGTM do you mind quickly scanning to see if TTL is used elsewhere and that it handles -1 and -2? (From looking at the TTL method). It may make sense to have a helper method:

```python
_ttl(client, ...) -> Optional[int]:
    # code in TTL method
```

Which would probably be:
https://github.com/jazzband/django-redis/blob/b1a68e05dfccf462cb5538e239ea78ba26067811/django_redis/client/default.py#L571-L581

Not sure we'd need to check for key existence like the TTL method does

terencehonles:
> No worries it is just mypy coverage that gets fooled by comments

We can probably configure the mypy target so it's more forgiving. I'm guessing it's actually that you've added lines (possibly even a single) that's using either `Any` or is an unchecked function. Either way a minor drop isn't a big deal. Having the reporter configured was mainly to encourage typing and allow us to monitor drops, but if we're ok with it were ok with it. I'm surprised it's so strict by default (I figured it would at least allow less than a percent change)

WisdomPill:
Sure I can change that with `self.ttl` as it was mentioned in an issue and fix also the other PR

WisdomPill:
Just noticed that version and client were not tested before and fixed another bug, should I add tests for them?

terencehonles:
> Just noticed that version and client were not tested before and fixed another bug, should I add tests for them?

If it's simple to do so it may make sense to do it now so we don't forget

django_redis/client/sharded.py
```diff
@@ -198,6 +198,12 @@ def incr_version(self, key, delta=1, version=None, client=None):
 
         try:
             ttl = client.ttl(old_key)
+
+            # redis ttl returns -1 if key does not have a timeout,
+            # so ttl has to be set to None to preserve the absence of a timeout
+            # in the new version as well
+            if ttl == -1:
```
terencehonles, line 205: This should be the same thing as above

WisdomPill:
I just noticed that since client was used and key constructed with version what I thought was a bug it is not happening so no bug, no need to add tests for something working correctly

django_redis/client/default.py
```diff
@@ -205,7 +205,7 @@ def incr_version(
         value = self.get(old_key, version=version, client=client)
 
         try:
-            ttl = client.ttl(old_key)
+            ttl = self.ttl(old_key, version=version, client=client)
```
terencehonles, line 208: I assume make_key isn't expensive? (it's going to be called twice now). The final comment I have is TTL could still be returned as 0. That's ok for this function, right? Not sure if that needs to be handled at all (I'm assuming it is, but just calling it out)

WisdomPill: it needs to be called because the old key has to be deleted

WisdomPill: line 221

django_redis/client/sharded.py
WisdomPill: You are right, nice catch!

django_redis/client/default.py
terencehonles: Ack, I understand what this does better, thanks

WisdomPill:
By the #499 is a different issue yet similar, that is why I created another PR, I know they probably could have been closed together in the same PR, it is just for the sake of following habits

terencehonles:
> By the #499 is a different issue yet similar, that is why I created another PR, I know they probably could have been closed together in the same PR, it is just for the sake of following habits

We should just close them together since they're affecting the same code path. Sorry I didn't realize they were different (they looked pretty similar) so I linked them both to close with this PR

terencehonles:
Oh sorry, is that a different incr method? It may make sense to just pull them in, but if you want them separate that's fine with me

WisdomPill:
No problem, let's just get going with closing this issues and maybe another before writing the CHANGELOG so that we could release a new version

terencehonles:
Oh sorry, is that a different incr method? It may make sense to just pull them in, but if you want them separate that's fine with me

> No problem, let's just get going with closing this issues and maybe another before writing the CHANGELOG so that we could release a new version

Well we still don't have access, right? 😅

But we should try to make sure we're updating the change log otherwise it will be more annoying to do after the fact

WisdomPill:
Let's keep them separate

terencehonles:
> Let's keep them separate

You can remove the fixes line in the PR description then

WisdomPill:
I will take care of the CHANGELOG before releasing, I already created the issue, I just would like to close a few issues this weekend before declaring a new version

WisdomPill:
> > Let's keep them separate
> 
> You can remove the fixes comment in the PR description then

Nice! I did not use the fixes in the comment for closing issues, very interesting an elegant way of doing that

WisdomPill:
Okay I am stupid i added the sharded edit in this PR. Let me fix that real quick

PR #524, opened 2021-05-01
Fixed incr when number is larger than 64 bit and has no timeout

### LINKED ISSUE #483: incr_version and decr_version delete keys with infinite timeout

WisdomPill:
Very similar to #523 it is applied to the `.incr` and consequentially to the `.decr` method.

Probably there will be the same mypy coverage issue as the other PR

fixes: #499

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -507,13 +507,7 @@ class DefaultClient:
                 # In this situations redis will throw ResponseError
 
                 # try to keep TTL of key
-
-                timeout = client.ttl(key)
-
-                # returns -1 if the key does exist and does not have expiration
-                # keeping it that way by setting timeout to None
-                if timeout == -1:
-                    timeout = None
+                timeout = self.ttl(key, version=version, client=client)
 
                 # returns -2 if the key does not exist
                 # means, that key have expired
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -509,6 +509,12 @@ class DefaultClient:
                 # try to keep TTL of key
 
                 timeout = client.ttl(key)
+
+                # returns -1 if the key does exist and does not have expiration
+                # keeping it that way by setting timeout to None
+                if timeout == -1:
+                    timeout = None
+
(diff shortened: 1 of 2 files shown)


terencehonles:
Is this the same as the other now?

django_redis/client/default.py
```diff
@@ -507,8 +507,8 @@ def _incr(
                 # In this situations redis will throw ResponseError
 
                 # try to keep TTL of key
+                timeout = self.ttl(key, version=version, client=client)
```
terencehonles, line 510: Does the sharded client need this too or does it use this function?

terencehonles (approved):

django_redis/client/default.py
WisdomPill: It uses this function

WisdomPill:
> Is this the same as the other now?

Nope, this is about `.incr` method the other about `.incr_version`. If the changes are okay with you, you can just squash and merge one of the two issues and I will solve any merge conflicts if they arise

terencehonles:
You can squash them, I'm fine with both

WisdomPill:
I did not understand very well, you mean to push a commit squaring all the commits or to use `squash and merge` button in this PR? I usually just use the button in the pr on normal days of work

terencehonles:
I'm just saying I'll let you do what you do 😅

terencehonles:
There probably won't be conflicts so it won't matter how you do it

PR #525, opened 2021-05-03
tweak codecov.io settings

terencehonles:
I believe this should be sufficient to make mypy less sensitive (it will do both status checks, but I think this is probably fine for either). We can also set a [threshold](https://docs.codecov.io/docs/commit-status#threshold) if you think we should allow mypy coverage dropping @WisdomPill (that one is per check).

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,4 +1,8 @@
 coverage:
+  # small changes should not fail coverage
+  precision: 1
+  round: up
+
   status:
     project:
       default:


.github/codecov.yml
```diff
@@ -1,4 +1,8 @@
 coverage:
+  # small changes should not fail coverage
+  precision: 1
+  round: up
```
WisdomPill, line 4: What does this mean?

WisdomPill:
LGTM by the way

.github/codecov.yml
terencehonles: It makes codecov only use one digit of precision (default 2) and when it does any rounding (it always will) it will round up (default down). You'll notice the precision change in the comment above:

> Flag	Coverage Δ	
> mypy	`33.7% <ø> (ø)`	
> tests	`81.8% <ø> (ø)`	

vs before:

> Flag	Coverage Δ	
> mypy	`33.61% <0.00%> (+0.09%)`	
> tests	`81.76% <100.00%> (ø)`	

The reason I chose up instead of down, is that if coverage drops by .01 percent it will only fail if it drops a full 10th of a percent (or at least that's what I understand or am trying to achieve). We can also set the threshold as mentioned in the description if this is not sufficient.

PR #528, opened 2021-05-06
Add timeout to prevent stuck actions

jackton1:
Resolves:

Screen Shot 2021-05-06 at 3 29 01 PM

Been running for 4hrs+

@niwinz @jdufresne @terencehonles

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ jobs:
       Django ${{ matrix.django-version }},
       Redis.py ${{ matrix.redis-version }}
 
+    timeout-minutes: 60
     runs-on: ubuntu-latest
 
     strategy:
@@ -98,6 +99,7 @@ jobs:
   lint:
     name: Lint (${{ matrix.tool }})
     runs-on: ubuntu-latest
+    timeout-minutes: 60
 
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       Django ${{ matrix.django-version }},
       Redis.py ${{ matrix.redis-version }}
 
-    timeout-minutes: 60
+    timeout-minutes: 10
     runs-on: ubuntu-latest
 
     strategy:
@@ -99,7 +99,7 @@ jobs:
   lint:
     name: Lint (${{ matrix.tool }})
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 10
 
     strategy:
       fail-fast: false


terencehonles:
Do you have an example of where this is stuck? We probably want it lower than 60m. I'm seeing the default is 360m.

terencehonles:
Ack, let's wait the 30m more to see if there's any error output, but we probably can lower it to 5m or 10m. We _shouldn't_ have to do either though. Was there any other times you've seen this?

jackton1:
It might be rare enough that we can ignore it. I know I've had to enable it on some repo's as a safe gaurd.

jackton1:
> Ack, let's wait the 30m more to see if there's any error output, but we probably can lower it to 5m or 10m. We _shouldn't_ have to do either though. Was there any other times you've seen this?

Seems it's still running.

terencehonles:
Hmm, well are we sure this option would even be respected in this case? It's hit the 6hr cap for jobs and it should have died at that point. Setting it lower likely would have been ignored :thinking: 

Where else have you seen this? Definitely surprised you caught this within 6hrs :sweat_smile:

terencehonles:
It's possible it's this https://www.githubstatus.com/incidents/nqyjr7kspl45

jackton1:
Hit 20+hrs wonder why they don't auto kill actions that have exceeded the max timeout.

Screen Shot 2021-05-07 at 7 00 09 AM

WisdomPill:
Does this PR still have a reason to exist? Is it feasible?

terencehonles:
It looks like they might have terminated it after 72 hours (I was wondering what would happen), and that seems to be based on [the workflow limit](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#:~:text=workflow%20run%20time%20-%20each%20workflow%20run%20is%20limited%20to%2072%20hours.%20if%20a%20workflow%20run%20reaches%20this%20limit%2C%20the%20workflow%20run%20is%20cancelled.). The issue we're seeing where it lasting longer than 6 hours seems to be mentioned in the community here: https://github.community/t/timeout-minutes-being-ignored/17847.

We don't have permissions to [see if this counted as billable time](https://docs.github.com/en/actions/managing-workflow-runs/viewing-job-execution-time) (which is really all I "care" about). If something got stuck and no one got charged and nothing was running it's not a big deal.

I'm _assuming_, but I may not be correct that this was related to the incident https://www.githubstatus.com/incidents/nqyjr7kspl45 and that setting a job limit would have been ignored because the default of 6 hours was also ignored. However, @WisdomPill I'm willing to allow you to be the tie breaker. The patch will lower the timeout to a much lower value which is possibly a good idea, but it may not resolve the issue that we saw. Feel free to approve the run to double check the workflow syntax and merge if you think we should add the change :slightly_smiling_face: .

WisdomPill:
seems fair, for safety it could be a nice to have

jackton1:
I'm fine closing this for now.  

Since there's no clear reason why this should happen.

PR #529, opened 2021-05-20
Add a way to provide the sentinel master name in the configuration

j-bernard:
In the current version, the master name is the redis url host.
This PR kept it as the default but a SERVICE_NAME configuration  parameter allows specifying it.

The Sentinel part of the readme was also fixed and updated but the code block is still not displayed in the github main page.
It seems to be related to the caption option as some other code blocks with caption are not displayed either.


WisdomPill:
I am leaving this to @terencehonles as he's more familiar with sentinel and he is also the one who kindly added support for it.

terencehonles:
I'm not sure what this buys since the service name is already defined by the hostname, and maybe it needs to be more clear that is the case rather than adding another option. Why might you need two different values? I don't believe the hostname would be used anywhere if you made this change.

j-bernard:
You're right, thanks.

I pushed with a comment to clarify it and kept my other commit fixing the doc.

WisdomPill:
@terencehonles friendly ping, just to remind you to review this

PR #531, opened 2021-05-26
Added changelog for the new version

WisdomPill:
I thought that maybe it is time for a major release because there are a lot of changes.

We might also not do it, it is up for debate.

I did not include #508 but it might be merged before the new release, let's see how it goes

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,24 @@
 Changelog
 =========
 
+Version 5.0.0
+--------------
+
+Date: TBD
+
+- supporting django 3.1 and django 3.2
+- dropped support for python 3.5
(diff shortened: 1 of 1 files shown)


terencehonles (approved):
LGTM, we may want to have people add their own entries and use [towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/) in the future.

WisdomPill:
Seems like a promising tool, you may want to add the issue and mark as a good frosty issue?

terencehonles:
Done https://github.com/jazzband/django-redis/issues/535

WisdomPill:
Thanks! Everything clear

Issue #532, opened 2021-05-26
Tests are failing in django master


WisdomPill:
Tests are failing for django master, can be seen in #531

WisdomPill:
Seems to be related to `tox`, @terencehonles do you have any idea on how to solve it?

neumantm:
Could
`django.utils.deprecation.RemovedInDjango50Warning: The default value of USE_TZ will change from False to True in Django 5.0. Set USE_TZ to False in your project settings if you want to keep the current default behavior.`
be the cause?

WisdomPill:
No, that's a warning there is also an `InvocationError` probably from `tox`

terencehonles:
The invocation error comes because the underlying command failed. It's the USE_TZ that's the underlying issue. I created #533 to fix the issue.

PR #533, opened 2021-05-26
Draft: Fix test failure caused by django deprecation

neumantm:
The failure was caused by:
django.utils.deprecation.RemovedInDjango50Warning:
The default value of USE_TZ will change from False to True in Django 5.0.
Set USE_TZ to False in your project settings if you want to keep
the current default behavior.

### DIFF
--- a/tests/settings/sqlite.py
+++ b/tests/settings/sqlite.py
@@ -1,4 +1,6 @@
 SECRET_KEY = "django_tests_secret_key"
+USE_TZ = False
+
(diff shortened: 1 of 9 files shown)


neumantm:
Possible fix for #532
I'm not sure though.
Don't have local dev env setup to run tests.

terencehonles:
Sorry I should paid more attention, thanks for fixing this. Duplicated by #534

PR #536, opened 2021-05-28
Added python 3.10 beta 1 to CI

WisdomPill:
Just so if there will be any issues we can tackle them and be ready for when it will be out of beta

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,7 @@ jobs:
           - '3.7'
           - '3.8'
           - '3.9'
+          - '3.10'
         django-version:
           - '2.2'
           - '3.1'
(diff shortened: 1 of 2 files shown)


WisdomPill:
I know, I changed the DeprecationWarning handling but it was breaking the test suite for python 3.10, if you come up with a more elegant solution I am all ears. Probably there is a better solution, I did not look into it that much

terencehonles (approved):
LGTM, I made some minor adjustments that I think you'd be fine with. I'll let you review and squash if you approve.

.github/workflows/ci.yml
```diff
@@ -31,17 +32,17 @@ jobs:
           # latest Django with pre-release redis
           - django-version: '3.2'
             redis-version: 'master'
-            python-version: '3.9'
```
terencehonles, line 34: Bumped Python version instead of adding more jobs

setup.cfg
```diff
@@ -66,17 +67,18 @@ envlist =
     isort
     mypy
     # tests against released versions
-    py{36,37,38,39}-dj{22,31,32}-redislatest
+    py{36,37,38,39,310}-dj{22,31,32}-redislatest
     # tests against unreleased versions
-    py39-dj32-redismaster
-    py39-djmain-redis{latest,master}
+    py310-dj32-redismaster
```
terencehonles, line 72: Bumped Python version instead of adding more jobs

Issue #539, opened 2021-06-03
Add issue template


WisdomPill:
Many times people open issues for issues not related to bugs but for knowing how to use the library, I suggest to add some disclaimers in the issue template for looking for answers elsewhere, like stackoverflow, and also add tags to the issues automatically, like question, bug and so on.

PR #545, opened 2021-08-24
created  issue templates #539

sarathak:
created  issue templates #539


WisdomPill:
wait, open source does not work like that, when you do stuff somebody else should merge and not do it by yourself.

sarathak:
@WisdomPill your correct 

i just clicked to see how it preventing but got merged 

i think there must have some rule that prevent to do so

Issue #535, opened 2021-05-27
Use towncrier to generate release notes


terencehonles:
This issue is to investigate how much config is needed to configure [`towncrier`](https://towncrier.readthedocs.io/en/actual-freaking-docs/) or a similar tool to generate the release notes instead of manually generating them.

An investigation in the form of a PR will be evaluated for its complexity, and upon approval will become part of the PR process, shifting the burden of release notes to happen during the PR review (this could still be added/adjusted by a maintainer instead of the original author, but at least the notes happen when the PR happens).

PR #546, opened 2021-08-28
Use towncrier to generate release notes #535

sarathak:
towncrier to generate release notes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,67 @@
+[build-system]
+requires = ["setuptools>=40.6.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+
+[tool.coverage.run]
+parallel = true
+branch = true
+source = ["django_redis"]
+
+[tool.coverage.paths]
+source = ["src", ".tox/*/site-packages"]
+
+[tool.coverage.report]
+show_missing = true
+exclude_lines = [
+    "pragma: no cover",
+    # PyPy is unacceptably slow under coverage.
+    "if PYPY:",
+]
+
+
+[tool.black]
+line-length = 79
+extend-exclude = '''
+# Exclude pattern matching test till black gains Python 3.10 support
+.*test_pattern_matching.*
+'''
+
+
+[tool.interrogate]
+verbose = 2
+fail-under = 100
+whitelist-regex = ["test_.*"]
+
+
+[tool.isort]
+profile = "attrs"
+
+
+[tool.towncrier]
+    package = "django_redis"
+    package_dir = "."
+    filename = "CHANGELOG.rst"
+    template = "changelog.d/towncrier_template.rst"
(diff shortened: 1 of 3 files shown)


WisdomPill:
I am not familiar with the tool @terencehonles is the one who proposed it, so I guess he is more comfortable with it than me

setup.cfg
```diff
@@ -174,3 +174,9 @@ ignore_missing_imports = true
 
 [mypy-redis.sentinel]
 ignore_missing_imports = true
+
+[testenv:changelog]
+basepython = python3.8
+deps = towncrier<21.3
```
terencehonles, line 180: why < 21.3? If there's a good reason do you mind putting your response in a comment so it's known for future editors?

setup.cfg
```diff
@@ -174,3 +174,9 @@ ignore_missing_imports = true
 
 [mypy-redis.sentinel]
 ignore_missing_imports = true
+
+[testenv:changelog]
+basepython = python3.8
```
terencehonles, line 179: I think this would be better to not fix this to 3.8 since any Python3 that `django-redis` supports should work.

```suggestion
basepython = python3
```

pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+
+
+[tool.interrogate]
+verbose = 2
+fail-under = 100
+whitelist-regex = ["test_.*"]
+
+
+[tool.isort]
+profile = "attrs"
```
terencehonles, line 38: Can you limit these changes to `towncrier`? or use `towncrier.toml` until we're ready to transition tools to use `pyproject.toml`?

pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+
+
+[tool.towncrier]
+    package = "django_redis"
```
terencehonles, line 42: I don't believe these need to be indented, and this doesn't match your indentation above nor do I believe it's common to indent tables in toml (I could be wrong about what's common).

terencehonles (commented):
@WisdomPill / @sarathak do you know if all of our tools support `pyproject.toml` now? There's config for existing tools that already is defined in `setup.cfg` and now there are two configs which will make it confusing to know which to edit. This PR should limit it's changes to just things that will affect `towncrier` and we should figure out if we're making the `pyproject.toml` migration now and do that via a separate PR.

I added some comments about the PR itself, and I'd need to test the setup, but it seems like everything makes sense to me. One more thing that we'd want to include in this PR is some documentation changes letting PR authors know that the repo uses town crier and a quick description of how to use it (potentially linking to external documentation as necessary)

sarathak:
@terencehonles 
i have fixed all requested item. please check

terencehonles:
Thanks for the updates, I probably need a few days to get back to you, but I will keep this notification unread so I can come back to it

PR #551, opened 2021-09-26
feature: Zstandard compressor

vbudovski:


WisdomPill (approved):

WisdomPill:
thanks for the PR @vbudovski, just the mypy error is needed to be solved, redis master is broken for a while now, I will have a look at that and solve it in another issue

Issue #521, opened 2021-04-08
Convert unittest class base tests to pytest function tests


terencehonles:
By converting unittest class base tests to pytest function based tests we can use some nifty features of pytest such as fixtures* and parameterization :tada: 

See: https://docs.pytest.org/en/stable/unittest.html#pytest-features-in-unittest-testcase-subclasses

Example of what needs to be done: #520

WisdomPill:
It has been merged now, as a matter of fact we could split `test_backend.py` unittest classes in it's own module (namely `backend`) to have smaller files, separate testing different aspects of the library and proceed by steps.

I think a test of one part of backend might simplify contributing.

@terencehonles do you agree?

terencehonles:
Definitely, the classes may make sense to be how things are split up (I haven't looked at how they are organized), and if someone needs to share fixtures between tests they can add the file `conftest.py` in the tests package and put the fixtures there. When placed there they will be requestable by any modules at or below that level.

joonyoungleeduke:
I can try working on this!

PR #553, opened 2021-10-09
#521, Convert unittest class base tests to pytest function tests

rootart:
Many thanks for the great package folks 🤗! 
I managed to convert the majority of tests from unittest class base tests to pytest and will appreciate your feedback and reviews. 
`SessionTests` tests were not migrated as they look to be very much coupled with ones from Django.


tests/test_backend.py
```diff
@@ -31,174 +31,166 @@ def reverse_key(key):
...
-        res = self.cache.get("test_key_nx")
-        self.assertIsNone(res)
+        default_cache.delete("test_key_nx")
+        res = default_cache.get("test_key_nx")
+        assert res is None
 
-        res = self.cache.set("test_key_nx", 1, nx=True)
-        self.assertTrue(res)
+        res = default_cache.set("test_key_nx", 1, nx=True)
+        assert res
```
WisdomPill, line 128: I prefer it to be more specific
```suggestion
        assert res == True
```

tests/test_backend.py
```diff
@@ -31,174 +31,166 @@ def reverse_key(key):
...
+def test_connection_strings(connection_string):
+    cf = pool.get_connection_factory(
+        path="django_redis.pool.ConnectionFactory", options={}
+    )
+    res = cf.make_connection_params(connection_string)
+    assert res["url"] == connection_string
+
+
+@pytest.fixture()
+def available_caches(settings):
```
WisdomPill, line 51: could you add a little bit of type hints? and maybe also separate them? just so that it looks cleaner and you do not have to do the unpacking in every test

tests/test_backend.py
```diff
@@ -31,174 +31,166 @@ def reverse_key(key):
...
 
-        for key in ["foo-aa", "foo-ab", "foo-bb", "foo-bc"]:
-            self.cache.set(key, "foo")
+    for key in ["foo-aa", "foo-ab", "foo-bb", "foo-bc"]:
+        cache.set(key, "foo")
 
-        res = self.cache.delete_pattern("*foo-a*")
-        self.assertTrue(bool(res))
+    res = cache.delete_pattern("*foo-a*")
+    assert bool(res)
```
WisdomPill, line 103: better to check if it is True
```suggestion
    assert res is True
```

tests/test_backend.py
```diff
@@ -31,174 +31,166 @@ def reverse_key(key):
...
+        res = default_cache.get("test_key_nx")
+        assert res is None
 
         # test that timeout will not affect key, if it was there
-        self.cache.set("test_key_nx", 1)
-        res = self.cache.set("test_key_nx", 2, timeout=2, nx=True)
-        self.assertFalse(res)
+        default_cache.set("test_key_nx", 1)
+        res = default_cache.set("test_key_nx", 2, timeout=2, nx=True)
+        assert not res
```
WisdomPill, line 150:
```suggestion
        assert res is False
```

tests/test_backend.py
```diff
@@ -208,534 +200,536 @@ def test_save_dict(self):
...
 
-        res = self.cache.get_many(["a", "b", "c"])
-        self.assertEqual(res, {"b": 2, "c": 3})
+        res = default_cache.get_many(["a", "b", "c"])
+        assert res == {"b": 2, "c": 3}
 
-        res = self.cache.delete("a")
-        self.assertFalse(bool(res))
+        res = default_cache.delete("a")
+        assert not bool(res)
```
WisdomPill, line 342:
```suggestion
        assert res is False
```

tests/test_backend.py
```diff
@@ -208,534 +200,536 @@ def test_save_dict(self):
...
             )
 
-    def test_delete(self):
-        self.cache.set_many({"a": 1, "b": 2, "c": 3})
-        res = self.cache.delete("a")
-        self.assertTrue(bool(res))
+    def test_delete(self, default_cache):
+        default_cache.set_many({"a": 1, "b": 2, "c": 3})
+        res = default_cache.delete("a")
+        assert bool(res)
```
WisdomPill, line 336:
```suggestion
        assert res is True
```

tests/test_backend.py
```diff
@@ -208,534 +200,536 @@ def test_save_dict(self):
...
-        self.assertEqual(type(res), bool)
-        self.assertFalse(res)
-
-    @patch("django_redis.cache.DJANGO_VERSION", (3, 0, 1, "final", 0))
-    def test_delete_return_value_type_before31(self):
+        mocker.patch("django_redis.cache.DJANGO_VERSION", new=(3, 1, 0, "final", 0))
+        default_cache.set("a", 1)
+        res = default_cache.delete("a")
+        assert type(res) == bool
+        assert res
```
WisdomPill, line 350:
```suggestion
        assert res is True
```

tests/test_backend.py
```diff
@@ -208,534 +200,536 @@ def test_save_dict(self):
...
+        res = default_cache.delete("b")
+        assert type(res) == bool
+        assert not res
```
WisdomPill, line 353:
```suggestion
        assert res is False
```

tests/test_backend.py
```diff
@@ -208,534 +200,536 @@ def test_save_dict(self):
...
-        res = self.cache.delete("b")
-        self.assertEqual(type(res), bool)
-        self.assertFalse(res)
-
-    @patch("django_redis.cache.DJANGO_VERSION", (3, 0, 1, "final", 0))
-    def test_delete_return_value_type_before31(self):
+        mocker.patch("django_redis.cache.DJANGO_VERSION", new=(3, 1, 0, "final", 0))
+        default_cache.set("a", 1)
+        res = default_cache.delete("a")
+        assert type(res) == bool
```
WisdomPill, line 349: let's keep isinstance for simplicity and consistency 
```suggestion
        assert isinstance(res, bool)
```

tests/test_backend.py
```diff
@@ -208,534 +200,536 @@ def test_save_dict(self):
...
+        assert res
+        res = default_cache.delete("b")
+        assert type(res) == bool
```
WisdomPill, line 352: let's keep isinstance for simplicity and consistency 
```suggestion
        assert isinstance(res, bool)
```

tests/test_backend.py
```diff
@@ -208,534 +200,536 @@ def test_save_dict(self):
...
+        assert type(res) == int
+        assert res == 1
+        res = default_cache.delete("b")
+        assert type(res) == int
+        assert res == 0
+
+    def test_delete_many(self, default_cache):
+        default_cache.set_many({"a": 1, "b": 2, "c": 3})
+        res = default_cache.delete_many(["a", "b"])
+        assert bool(res)
```
WisdomPill, line 369:
```suggestion
        assert res is True
```

tests/test_backend.py
```diff
@@ -208,534 +200,536 @@ def test_save_dict(self):
...
+
+        res = default_cache.get_many(["a", "b", "c"])
+        assert res == {"c": 3}
+
+        res = default_cache.delete_many(["a", "b"])
+        assert not bool(res)
```
WisdomPill, line 375:
```suggestion
        assert res is False
```

WisdomPill (changes requested):
Great job and thank you for contributing. We were talking in the past too separate classes in different files, could you separate tests that were in the same class in different files?

I did not review everything because it seems that the same suggestions will be made for the rest of the file and I will need anyway to review everything after the splitting and small fixes will be made

rootart:
Hi @WisdomPill and thank you for the review. 
>  Great job and thank you for contributing. We were talking in the past too separate classes in different files, could you separate tests that were in the same class in different files?

Sure, I will work on it. 

> I did not review everything because it seems that the same suggestions will be made for the rest of the file and I will need anyway to review everything after the splitting and small fixes will be made

That totally makes sense and I also noticed that a lot of issues were related to the automatic unittest -> pytest migration tool that I used and I wasn't very careful reviewing results 😅

tests/test_backend.py
rootart: Agree, I've been using https://github.com/pytest-dev/unittest2pytest to convert larger test classes and didn't thoroughly check results.

tests/test_backend.py
rootart: Deletion of the multiple keys returns back the number of deleted objects and therefore I used ` assert bool(res) is True` to cover these cases.

terencehonles:
> @WisdomPill would be great if you can approve workflow execution to check tests bow

Done

Issue #547, opened 2021-09-11
Change pickle default version to Python default


ljodal:
**Problem Statement**
This library defaults to the latest pickle protocol. This does not match Python's default version in all cases and can make it harder to perform safe upgrades.

When upgrading Python it can in many cases be useful to stay compatible with previous versions, for example if there's a need to do a rollback or rolling deploys are used. Using the latest pickle version cannot always be used in these situations. For example Python 3.7 supports up to pickle protocol 4, while Python 3.8 supports up to pickle protocol 5. Python itself handles backwards compatibility by not changing the default version until the next release when introducing a new protocol. Python 3.8 defaults to protocol version 4, even though it supports 5.

**Describe the solution you'd like**
I suggest changing the default pickle protocol from `-1` (ie the latest version) to `pickle.DEFAULT_PROTOCOL`. This will ensure that this library uses the same default as the current Python version and no special care will not be needed to ensure safe upgrades can be performed using this library.

**Describe alternatives you've considered**
The default pickle protocol used by this library can be set to the default through the settings (`pickle.DEFAULT_PROTOCOL`).

WisdomPill:
Hello @ljodal,

nice catch, would you be interested in providing a PR to solve this issue?

ljodal:
Yeah, I can do that :)

PR #555, opened 2021-10-14
Change pickle default version to Python default instead of highest version

rootart:
Fixes: #547 regarding using `pickle.DEFAULT_PROTOCOL` as the default pickle version.

### DIFF
--- a/tests/test_serializers.py
+++ b/tests/test_serializers.py
@@ -1,0 +1,26 @@
+import pickle
+
+import pytest
+from django.core.exceptions import ImproperlyConfigured
+
+from django_redis.serializers.pickle import PickleSerializer
+
+
+class TestPickleSerializer:
+    def test_invalid_pickle_version_provided(self):
+        with pytest.raises(
+            ImproperlyConfigured, match="PICKLE_VERSION value must be an integer"
+        ):
+            PickleSerializer({"PICKLE_VERSION": "not-an-integer"})
+
+    def test_setup_pickle_version_not_explicitly_specified(self):
+        serializer = PickleSerializer({})
+        assert serializer._pickle_version == pickle.DEFAULT_PROTOCOL
+
+    def test_setup_pickle_version_too_high(self):
+        with pytest.raises(
+            ImproperlyConfigured,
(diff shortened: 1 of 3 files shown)


tests/test_serializers.py
```diff
@@ -0,0 +1,39 @@
...
+from django_redis.serializers.pickle import PickleSerializer
+
+
+class TestPickleSerializer:
+    @pytest.mark.skipif(
+        sys.version_info < (3, 8),
+        reason="Protocol 5 was added as highest from python version 3.8",
+    )
+    def test_using_negative_protocol_value(self):
+        serializer_p5 = PickleSerializer({"PICKLE_VERSION": pickle.HIGHEST_PROTOCOL})
```
WisdomPill, line 16: better to specify version 5 so that it will keep passing even when there will an eventual version 6 in the future
```suggestion
        serializer_p5 = PickleSerializer({"PICKLE_VERSION": 5})
```

terencehonles: -1 is always the highest protocol, but I'm a little confused why there's a skip for < Python 3.8. This should be valid for all Python versions and it's fine to just run the test everywhere instead of special casing this.

tests/test_serializers.py
```diff
@@ -0,0 +1,39 @@
...
+from django.core.exceptions import ImproperlyConfigured
+
+from django_redis.serializers.pickle import PickleSerializer
+
+
+class TestPickleSerializer:
+    @pytest.mark.skipif(
+        sys.version_info < (3, 8),
+        reason="Protocol 5 was added as highest from python version 3.8",
+    )
```
terencehonles, line 14:
```suggestion
```

tests/test_serializers.py
WisdomPill: I missed it, you are right, but then the assertion should be on the pickle_version used or the test does not make sense

terencehonles: Yeah, I agree. I'm fine removing the test since it's actually just tesitng that -1 is interpreted as `HIGEST_VERSION` by `pickle` and that's outside the domain of this package.

rootart:
> By the way, thank you very much for your contribution @rootart!!
@WisdomPill  it is my pleasure :) thank you for the reviews!

PR #557, opened 2021-10-19
configure pre-commit so pre-commit.ci stops complaining

terencehonles:
Just testing with some config that roughly matches the tox setup.


terencehonles:
@WisdomPill does this look good to you?

Issue #564, opened 2021-10-29
Default client does not return boolean value of the `expire` action of redis


puittenbroek:
**Describe the bug**
`expire` does not return the redis client expire value, which returns `True` if the key was actually in redis.
Or `False` if it wasn't present.
This means we have to manually check if they key is (or was) there after calling expire.

**To Reproduce**
Reproduced via shell_plus:
```python
    In [43]: from django.core.cache import cache
    In [47]: default_client = cache.client
    In [48]: redis_client = cache.client.get_client(write=True)
    In [49]: cache.expire('notakey', 9000) # no output since it's None
    In [50]: redis_key = default_client.make_key('notakey')
    In [52]: redis_client.expire(redis_key, 9000) # returns desired boolean value
    Out[52]: False
    In [53]: cache.set('actualkey', 'some value')
    Out[53]: True
    In [54]: cache.ttl('actualkey')
    Out[54]: 295
    In [55]: cache.expire('actualkey', 9000) # no output since it's None
    In [56]: cache.ttl('actualkey')
    Out[56]: 8999
    In [57]: redis_key = default_client.make_key('actualkey')
    In [58]: redis_client.expire(redis_key, 900000) # returns the desired boolean value
    Out[58]: True
    In [59]: cache.ttl('actualkey')
    Out[59]: 899998
```

**Expected behavior**
The return value of `expire` is returned instead of dropped.

Code [here](https://github.com/jazzband/django-redis/blob/master/django_redis/client/default.py#L290-L291) can be simplified.
To not needing the if exists checks.
```python
    return client.expire(key, timeout)
```

**Stack trace**
If applicable, add stack trace to help explain your problem.

**Environment (please complete the following information):**
 - Python version: 3.8.12
 - Django Redis Version: bug present on `master`, but we use `4.8.0`
 - Django Version: 2.2.24
 - Redis Version: 3.2.12
 - redis-py Version: 2.10.6

Note: We use `rediscluster` which provides `StrictRedisCluster`, which is a subclass of `StrictRedis` from the redis package.

**Additional context**

PR #565, opened 2021-10-29
Make expire, pexpire, expireat and persist return the redis client value

puittenbroek:
While working on fixing #564 I also saw that `pexpire` and `expireat` and `persist` used the same construction.
Applied boy scout rule :)


puittenbroek:
Seems friday afternoon wasn't best time to adjust tests, fixed them now and added file for towncrier.

django_redis/client/default.py
```diff
@@ -266,46 +266,43 @@ def get(
...
             client = self.get_client(write=True)
 
         key = self.make_key(key, version=version)
 
-        if client.exists(key):
-            client.expire(key, timeout)
+        return client.expire(key, timeout)
 
-    def pexpire(self, key, timeout, version=None, client=None):
+    def pexpire(self, key, timeout, version=None, client=None) -> int:
```
WisdomPill, line 291: why does this return an integer?

`PERSIST`, `EXPIRE` and `PEXPIRE` all return 1 or 0, why only two of them return booleans?

puittenbroek: Good question, I noticed it but forgot to look into the why.
Did so now :) Looking at redis documentation they all return int.
But redis-py parses the response according to the existing in `RESPONSE_CALLBACK` as seen here:
https://github.com/redis/redis-py/blob/396e7323959b7a79d72602afee3ef648dc23adf9/redis/client.py#L619-L624

It appears that `PEXPIRE` is simply not listed there, resulting in the 'raw' integer response.

WisdomPill: maybe then parsing it to boolean is better so that we keep the same interface, I will raise the issue to `redis-py` team. Can you also add `pexpire_at` since you are at it?

changelog.d/564.bugfix
```diff
@@ -0,0 +1 @@
+Make expire, pexpire, expireat and persist return the redis client value
```
terencehonles, line 1: I would probably characterize this as a new feature for this library rather than a bugfix.

Also, I'm not sure, but it's possible towncrier may handle a feature and bugfix being the same issue number, but I would personally expect a PR to have one or the other defined (though I can imagine a case where that's not true)

puittenbroek: Tested towncrier locally, it handles two files for the same issue.
I added the pexire_at at request of @WisdomPill :)

README.rst
```diff
@@ -393,22 +393,47 @@ specify a new expiration timeout using the ``persist`` and ``expire`` methods:
...
+    >>> cache.pttl("foo")
+    3600900
+
 The ``pexpire`` method can be used to provide millisecond precision:
 
 .. code-block:: pycon
 
     >>> cache.set("foo", "bar", timeout=22)
     >>> cache.pexpire("foo", timeout=5500)
+    1
```
WisdomPill, line 436:
```suggestion
    True
```

WisdomPill (approved):

puittenbroek:
Really fixed ShardClient testing now, verified locally.
There are still some tests that fail, but they seem unrelated to my changes?

terencehonles:
> Really fixed ShardClient testing now, verified locally. There are still some tests that fail, but they seem unrelated to my changes?

If you're referring to the ones testing redis master that's a known issue, I approved the workflow run to check where things are at.

PR #567, opened 2021-11-07
Changed ci rule

WisdomPill:
Changed ci rule so that it runs only on pull request on master or push on master


terencehonles:
Are we burning through our credits? Not sure why we need to restrict this just to a single branch.

WisdomPill:
We are not burning credits, but running jobs twice is not elegant, since we are using only master.

WisdomPill:
This is to avoid running jobs for pull_request and for push.

PR #568, opened 2021-11-08
Update type comments to type annotations

terencehonles:


WisdomPill:
Why move to type comments? We are already supporting 3.6+ so it should work on all supported python versions

terencehonles:
> Why move to type comments? We are already supporting 3.6+ so it should work on all supported python versions

This *removes* type comments, because it is supported in 3.6+. The type comments are the older syntax.

WisdomPill:
I guess I am stupid, or it's just monday morning

terencehonles:
> I guess I am stupid, or it's just monday morning

Haha, no prob. Maybe not enough coffee yet? :coffee:

Issue #569, opened 2021-11-08
Support redis-py 4.0


terencehonles:
**Problem Statement**

The [Redis Python client is now in RC1](https://github.com/redis/redis-py/releases/tag/v4.0.0rc1) but introduces some breaking changes (specifically seen with the Sentinel configuration). This library should be updated to support the new changes ideally in a backwards compatiable manner.

WisdomPill:
there is already an issue for this, it is #542

terencehonles:
Well I think they're related, but is this the only thing that's broken in redis 4?

hartwork:
CC #575 #576

aleehedl:
Is there any estimate when this will be available in a release?

WisdomPill:
I can release in the next days, max Monday. 

Can I ask why you are looking for redis-py 4 as a dependency?

WisdomPill:
I would wait for #580 to be merged and then release

aleehedl:
> Can I ask why you are looking for redis-py 4 as a dependency?

Yeah, because redis-py 4 was already installed into a production system and I'm not sure how a downgrade would pan out.

WisdomPill:
I can say that the only known issue is with sentinel commands that are different in 4.0.1 and 4.0.2, the rest of the versions are supported, I am waiting for @terencehonles to review #580 to support django 4 and drop support for django 2.2

WisdomPill:
Will release tomorrow

Issue #542, opened 2021-08-23
Sentinel commands have migrated in redis main 


sarathak:
Redis pre release test throwing error in CI

- Test Python 3.9, Django 3.2, Redis.py master
- Test Python 3.9, Django 3.2 main, Redis.py master

```python
  self = Sentinel<sentinels=[127.0.0.1:26379]>, service_name = 'default_service'
  
      def discover_master(self, service_name):
          """
          Asks sentinel servers for the Redis master's address corresponding
          to the service labeled ``service_name``.
      
          Returns a pair (address, port) or raises MasterNotFoundError if no
          master is found.
          """
          for sentinel_no, sentinel in enumerate(self.sentinels):
              try:
  >               masters = sentinel.sentinel_masters()
  E               AttributeError: 'Redis' object has no attribute 'sentinel_masters'
  
  .tox/py39-dj32-redismaster/lib/python3.9/site-packages/redis/sentinel.py:211: AttributeError
  _____ ERROR at teardown of DjangoRedisCacheTestEscapePrefix.test_iter_keys _____
  
  self = <tests.test_backend.DjangoRedisCacheTestEscapePrefix testMethod=test_iter_keys>
  
      def tearDown(self):

```

WisdomPill:
hello @sarathak,
redis master is allowed to fail as there might be new unreleased changes that will break the library.
Have you had look at what this might be?
Are you open to make changes to solve this for newer version of `redis-py`?

sarathak:
@WisdomPill 

my opinion is we have to remove pre release test from CI

WisdomPill:
why one earth would you want to do that?

pre releases is added for this very reason, so that we know that before releasing a new version we have to solve some issues first to support the next version of redis

WisdomPill:
In this case, sentinel and cluster commands have moved to it's own mixins, have a look [here](https://github.com/andymccurdy/redis-py/pull/1534). New commands are found [here](https://github.com/andymccurdy/redis-py/blob/e19a76c58f2a998d86e51c5a2a0f1db37563efce/redis/commands.py#L2953)

chayim:
We're planning on releasing a new redis-py in the short-term. This is a **great catch** that your project has in its pipeline, I highly recommend maintaining these tests. Note: We haven't broken the API itself, but I'd like to go down the path of more restructuring in the mid-term.

WisdomPill:
@chayim thanks for the notice. 
[...]

PR #570, opened 2021-11-08
pin redis-py to 3.x until 4.x breaking changes can be addressed

terencehonles:


WisdomPill:
LGTM

WisdomPill:
Is it possibile to not check for coverage on tests directory but only check types in tests directory?

terencehonles:
> Is it possibile to not check for coverage on tests directory but only check types in tests directory?

Is that why it's codecov is complaining? I figured it was not enough typing over there. Let me look at the config.

terencehonles:
It should be excluding the tests:

https://github.com/jazzband/django-redis/blob/9cd2bbeb5039f013abcdd24c1486748676e4c312/setup.cfg#L46-L49

terencehonles:
I think what you're seeing is that Mypy is complaining about any `Any` which the codebase has a lot since anything that's not defined is an `Any`.

jacklinke:
Is there a description of the problem or an associated issue that details what issues `django-redis` has with `redis-py` 4?

terencehonles:
@jacklinke right now what's known is #542. I had created #569 which may have been more clear, but it was closed as a dup of #542.

I understand the problem, but haven't had time to adjust the code. It does seem like it will be a "decent" amount of work because redis 4 separates sentinel and non sentinel connection pools a bit more. This is _probably_ a good thing since you can configure the connection settings more independently, but not how I had written #460 and the code will need a little re-work. After that we would need to look for any other breaking changes and make sure we have sufficient test coverage to make sure expectations continue to be met regardless if breaking changes are documented or not.

WisdomPill:
@terencehonles regarding tests, 
lately I have been contributing a little in `redis-py` and I saw how they are using `tox` with docker for setting up the environment.
I find it very clean since you just need docker running and the test suite will take care of the rest.

Would you mind taking a look and get back to me?
I would like to have something similar in the future

terencehonles:
@WisdomPill do you mind creating a separate issue to discuss and research that? We probably don't need to keep bumping this thread with relation to a change like that.

PR #575, opened 2021-12-03
Restored redis 4 and master in CI

WisdomPill:


WisdomPill:
@hartwork do you think we could work around the issue with `packaging` by installing it when using `redis-master`?

hartwork:
> @hartwork do you think we could work around the issue with `packaging` by installing it when using `redis-master`?

@WisdomPill technically yes, but please note https://github.com/jazzband/django-redis/pull/576#issuecomment-992511729 for the bigger picture.

hartwork (approved):

hartwork (changes requested):
@WisdomPill Installation of package `packaging` seems still missing.
PS: My vote for merging this without squashing.

hartwork (changes requested):
@WisdomPill the way how `packaging` is added right now does not guarantee that it's installed _prior to_ redis-py, which is causing the breakage. If that's not easily fixable, my vote for re-opening #576 and merge that fix in isolation, first.

WisdomPill:
agreed

hartwork:
@WisdomPill thank you!

PR #576, opened 2021-12-03
Unblock redis>=4.0.2 (#542, #570)

### LINKED ISSUE #542: Sentinel commands have migrated in redis main

hartwork:
Fixes #542

Upstream commit https://github.com/redis/redis-py/commit/d2b233384458869270352b8c99ca682ae480da5f that is included with redis-py >=4.0.2 adds method `Redis.sentinel_masters` by making class `Redis` inherit from `SentinelCommands`, where the related code resides.

Related to issue #542, follow-up to pull request #570

### DIFF
--- a/changelog.d/542.bugfix
+++ b/changelog.d/542.bugfix
@@ -1,0 +1,1 @@
+Block use with broken redis-py 4.0.0 and 4.0.1
--- a/changelog.d/576.misc
+++ b/changelog.d/576.misc
@@ -1,0 +1,1 @@
+Unblock redis-py >=4.0.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,10 +38,10 @@ packages =
     django_redis.compressors
 install_requires =
     Django>=2.2
-    redis>=3,<4
+    redis>=3,!=4.0.0,!=4.0.1
 
 [options.extras_require]
-hiredis = redis[hiredis]>=3,<4
+hiredis = redis[hiredis]>=3,!=4.0.0,!=4.0.1
 
 [coverage:run]
 omit =


WisdomPill:
Sorry I away from my laptop, will have a look at it by the end of weekend

hartwork:
@WisdomPill thanks!

WisdomPill:
It is okay for me, but now I have some doubts about the matrix strategy, I think we should test with specific versions of redis-py and maybe also specific versions of redis server.

By the way there is a comment in the ci action which should be uncommented, like [here](https://github.com/jazzband/django-redis/pull/575/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f).

@terencehonles what do you think?

terencehonles:
@WisdomPill I pushed the commit from this PR to #575 because I both have permission to push to that branch, but also it is the original branch. The change seems fine to me, but I will not have a chance to test it for a few days because I'm flying now and will likely be busy for a bit. I don't know why this should be an issue since this should return the previous functionality back to how it was before. I do think there are some issues with the connection pools Redis configures since now there is both a sentinel pool and a redis pool, but that should have already been an issue previously and this change should likely be sufficient until we have a more pressing need to allow independent connection pool configuration.

WisdomPill:
Okay, let's finish this on #575 then, but I see that there is an issue with packaging and there is already an issue open in redis/redis-py#1784 about it, so I guess we can wait to see what happens, by the way @hartwork is already involved in that conversation.

hartwork:
> By the way there is a comment in the ci action which should be uncommented, like [here](https://github.com/jazzband/django-redis/pull/575/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f).

@WisdomPill I had to keep those commented because redis-py `master` is broken in that regard, which I know you know by now. While I personally like the idea of becoming aware of future issues early, unless the CI gets an "accept failure" flag for those moving targets, moving targets are by design a problem to CI stability and reproducibility, so maybe not testing against redis-py `master` is in the interest of everyone working on pull requests around, for now.

WisdomPill:
I see your point, but I think keeping master on the CI and ignoring it when there are known issues is vital, 
especially for the people maintaining those libraries, 
in our case if we did not know that redis-py 4 was breaking sentinel support then our users would have discovered it in a painful way.
I see it as our duty to inform maintainers of libraries we use that the next version has something wrong that should be addressed before releasing on our or their side. 
Even if it breaks and stays broken for sometime, we can always remove it and restore it.

hartwork:
> I see your point, but I think keeping master on the CI and ignoring it when there are known issues is vital, especially for the people maintaining those libraries

Yes, but as long as any single CI task fails the whole thing will marked as failed and then people learn that CI is often red and always needs extra checking.


> I see it as our duty to inform maintainers of libraries we use that the next version has something wrong that should be addressed before releasing on our or their side. Even if it breaks and stays broken for sometime, we can always remove it and restore it.

I see your point, but personally I think waiting for future breakage and bug reports and a fully green CI is more healthy. I guess it's okay that we disagree on that, it's just my 2 cents.

PR #580, opened 2021-12-17
Added django 4 to CI

### LINKED ISSUE #579: Add django 4 to CI
**Problem Statement**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

WisdomPill:
Fixes #579 

Added support for django 4.0


hartwork (commented):
Hi @WisdomPill , four things I noticed:
- Adding support for Django 4 seems uncontroversial, dropping 2.2 prior to EOL could be considered controversial by some; I'd split those in two pull requests personally
- The current title of the pull request does not mention dropping Django 2.2; my vote for adding that
- Django 3.1 is EOL even before Django 2.2 (see image below) but it is not being dropped; why not?
- Dropping of Django 2.2 seems to be incomplete, e.g. I still see a line `2.2: dj22` in `setup.cfg`, maybe more.

Just my 2 cents.

Django releases

hartwork:
> * Dropping of Django 2.2 seems to be incomplete, e.g. I still see a line `2.2: dj22` in `setup.cfg`, maybe more.

PS: @WisdomPill there is indeed one more: `README.rst` still says Django `2.2+`.

WisdomPill:
my mistake, dropping 2.2 is a mistake the docs were saying that support for django 1.11 has dropped in april 2020

hartwork (approved):

.github/workflows/ci.yml
```diff
@@ -40,12 +40,19 @@ jobs:
             redis-version: 'latest'
             python-version: '3.10'
 
-          # TODO: re-enable with https://github.com/jazzband/django-redis/issues/569
           # pre-release Django and redis
           - django-version: 'main'
             redis-version: 'master'
             python-version: '3.10'
 
+        # exclude python 3.6 and 3.7 for django 4.x
```
hartwork, line 48: Maybe a rationale on why those are excluded would make this perfect…

WisdomPill: They are not supported, maybe they are using new syntax or are looking forward to it. I will write something

hartwork: Thanks!

WisdomPill:
thanks @hartwork !

PR #583, opened 2021-12-23
Moving to tox-docker

### LINKED ISSUE #574: Adapting redis-py tox strategy
@terencehonles regarding tests,
lately I have been contributing a little to redis-py and I saw how they are using tox with docker for setting up the environment.
I find it very clean and elegant since you just need docker running and the test suite will take care of the rest.

Would you mind taking a look and get back to me?
I would like to have something similar in the future for django-redis as well.

Have a look [here](https://github.com/redis/redis-py/blob/175a05f4de17918b74bde7f554182968b1f6aabb/tox.ini)

WisdomPill:
Fixes #574 

This PR is to make tox manage docker as `redis-py` is doing.

I still have some issues with sentinel, I seem to not understand how it should work.

@chayim could you help me understand where I got the configuration wrong?
Also another couple of questions, why you are not using redislab docker hub repository?
What is the difference?

@terencehonles maybe you also have some hints?

After that the unix socket will be the last problem to solve, but for now I think it is not very important


.github/workflows/ci.yml
```diff
@@ -79,18 +79,10 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade tox tox-gh-actions
```
chayim, line 82: @WisdomPill You may want to think about getting these out of your github actions and into something like a dev_dependencies (or pyproject.toml dev area). This way they're not as hidden from users.

chayim:
@WisdomPill Awesome to see you also thinking of this approach! We ended up spinning our own dockers and pushing them to dockerhub, because it was the easiest way for us to proceed - while still having everything we need. Maybe that workflow is useful, or maybe those dockers are useful?

I'm planning on (future) adding support for multiple versions of redis as well, so maybe we an collaborate there? If it helps any, I'm happy to push to dockerfab for you.  Also - let me know how I can help!

.github/workflows/ci.yml
WisdomPill: Yes, I guess I will add some kind of dev script to setup the environment.

I still do not really understand how tox installs dependencies and where is it getting `redislatest`.

These are my first steps with tox

WisdomPill:
Now the problem seems to be that sentinel does find the master node, or that it is disconnected somehow.

Do you know if the configuration is wrong? 
The previous configuration scripts are under [test/start-redis.sh](https://github.com/jazzband/django-redis/blob/4e7f2d472929b317b3135f84ccd62d675e32a810/tests/start_redis.sh) and [test/wait_for_redis.sh](https://github.com/jazzband/django-redis/blob/4e7f2d472929b317b3135f84ccd62d675e32a810/tests/wait_for_redis.sh).
They were setup in the [this](https://github.com/jazzband/django-redis/blob/4e7f2d472929b317b3135f84ccd62d675e32a810/.github/workflows/ci.yml#L85) manner.
This is my first experience with sentinel, usually it is @terencehonles who is in charge of it.

I would love to keep contributing maybe even with some python code to redis-py, count me in for testing multiple versions of redis, this is something we are also looking for.

After solving this issue I would like also to tackle the usage of original docker images from redis.

.github/workflows/ci.yml
chayim: Got it. Here's how I think about tox with docker:

- Define a series of docker images that may or may not get used
- Define tox environments that depend on individual dockers
- Add pre_commands or dependencies if/as necessary
- Run tests

For [one project](https://github.com/RedisGears/rgsync/blob/master/tox.ini) I use this pattern to parameterize a pytest option, so that we can control test execution and dockers being run. It may be a better example for your pattern.

chayim:
> Do you know if the configuration is wrong?

Our approach is a bit different. I pushed a series of [dedicated dockers](https://github.com/redis/redis-py/tree/master/docker) that I maintain and built once, from the community redis docker to dockerhub. Those dockers each load a configuration file from disk. You'll note the [base docker](https://github.com/redis/redis-py/tree/master/docker/base) folder, and the subsequent configuration files for a  [sentinel node](https://github.com/redis/redis-py/blob/master/docker/sentinel_1/sentinel.conf), or a [replica](https://github.com/redis/redis-py/blob/master/docker/replica/redis.conf).

Extending that we have tox-docker run a [healthcheck](https://github.com/redis/redis-py/blob/1b7d5bbac89d1dc65154e52360af43373d45ebc6/tox.ini#L20) for each docker, ensuring they're at least up.  By controlling redis from outside, with the configuration file it becomes easy to (more) rapidly iterate on particular configuration needs.

This was the reason I built the [base docker](https://github.com/redis/redis-py/blob/master/docker/base/Dockerfile) from the [community docker image](https://hub.docker.com/_/redis), rather than our [redis enteprise docker](https://hub.docker.com/r/redislabs/redis). Using that docker, and loading a configuration file that is mounted into the container using the volume mounts, added the flexibility needed for environments.

Hope this helps @WisdomPill

WisdomPill:
@chayim thanks for the great explanation.

But, my problem is not on how to use tox-docker, my problem is with sentinel, it seems like the sentinel does not connect to the redis master, so my question is, is there something wrong with the configuration I have made? Could you have a look at it?

I copied the configuration from redis-py and tried to make it work, but for a normal instance is working, for sentinel not.

chayim:
@WisdomPill Can you help walking me through running this locally on a single python? Looking at the dockers you're starting via _tests/start_redis.sh_ the sentinel definitely sees the master - at least locally. I wasn't able to get pytest in my world functioning and running your tests - which I'm sure is on my side.

But, I validated connectivity by running (locally, outside of the dockers)
```
# connecting to the master, yes it's up
redis-cli -p 6379
PING # returns an ok
```

Now, I connected to the sentinel node
```
# to connect to the sentinel node
redis-cli -p 26379
PING # returns an ok

# to show me sentinel settings (since I see master0:name=default_service) I'm sure it is configured to see the master
INFO

# allows me to validate that yes, it is pointed to a master
SENTINEL get-master-addr-by-name default_service

# show me information from sentinel about the master
SENTINEL MASTER default_service
```

With the above ```INFO``` command, in the sentinel context you'll see _last-ok-ping-reply_ and it constantly changes, meaning there is connectivity. Similarly, running the redis-cli from within the individual dockers helped me validate connectivity another way.

I'm no sentinel expert, but I think you need [at least 3 nodes](https://redis.io/topics/sentinel) (so one more, see the text "Example 1: just two Sentinels, DON'T DO THIS). Because ```SENTINEL CKQUORUM default_service``` returns _NOQUORUM 1 usable Sentinels. Not enough available Sentinels to reach the specified quorum for this master_. Similarly, the other sentinel help commands ```SENTINEL replicas default_service``` and ```SENTINEL sentinels default_service``` both return empty arrays.

Hope this helps - and happy to debug more if you can help me on the environment side!

chayim:
@WisdomPill pinging you here too, as I'm happy to work on this in tandem.

WisdomPill:
I did not have time to work on open source lately, I will get back to you soon

terencehonles:
Sorry for ignoring you @WisdomPill I believe I saw the message at some point and meant to get back to you, but I obviously forgot. I think the docker compose implementation you landed on is a little better than this, or at least for the time being for its simplicity.

Looking at `tox-docker` it would tear down the containers for every tox invocation (and between the runs), this is good for database isolation, but to really benefit from this plugin we'd need to rework how the tests figure out what database to connect to.

If you want I can update the config to work for the current setup, but this has fallen out of date compared to the docker compose setup, and based on ^^^ I'm not sure it's worth putting the effort into.

PR #613, opened 2022-06-24
Log traceback if exceptions are ignored

### LINKED ISSUE #611: Add exception info to logger message when exceptions are ignored
We're using django-redis as a session cache backend for our Django project, with DJANGO_REDIS_IGNORE_EXCEPTIONS and DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS set to True. 

Unfortunately, we're experiencing a pesky Redis TimeoutError, which due to the previous settings, is only getting logged as:
```
ERROR    django_redis.cache 2022-06-08 17:16:54,470 cache.py _decorator 35 Redis TimeoutError: Timeout reading from socket
```

The issue happens only in the production environment, and we can't get a full exception traceback without impacting customers, so debugging this has been difficult. We'll probably work around this by live-editing the cache module on the web servers.

What are your thoughts on using `self.logger.exception` instead of `self.logger.error` in the [omit_exception decorator](https://github.com/jazzband/django-redis/blob/master/django_redis/cache.py#L35), or adding another setting (e.g. `DJANGO_REDIS_LOG_EXCEPTION_TRACEBACK`) which would control the logging behavior?

tmarice:
Fixes #611

### DIFF
--- a/changelog.d/611.misc
+++ b/changelog.d/611.misc
@@ -1,0 +1,1 @@
+Print full exception traceback when logging ignored exceptions
--- a/django_redis/cache.py
+++ b/django_redis/cache.py
@@ -32,7 +32,7 @@ def omit_exception(
         except ConnectionInterrupted as e:
             if self._ignore_exceptions:
                 if self._log_ignored_exceptions:
-                    self.logger.exception(str(e))
+                    self.logger.exception("Exception ignored")
 
                 return return_value
             raise e.__cause__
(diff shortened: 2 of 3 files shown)


WisdomPill (changes requested):

tmarice:
@WisdomPill codecov is complaining that the patch isn't covered with tests, so I'll add some that verify that the full traceback was logged.

I'm not sure the mypy failure is related to this PR, though.

tmarice:
@WisdomPill Do you mind taking another look at the PR?

WisdomPill:
sorry for being late, and very much thank you for the contribution

Issue #619, opened 2022-10-20
Why does `DJANGO_REDIS_CONNECTION_FACTORY` need to be a global django setting ?


vedantpuri:
## Context
We use django-redis for our project and I am in a situation where I need to make only one of my caches sentineled.
Ideally the others should also be converted but due to time constraints and other reasoning based on usage we want to convert only one for now. 

## Question
When trying to change this I noticed that this setting is global, which means that all my caches need to be sentineled for this to work. I am not exactly sure why this is global and/or if this could be moved into OPTIONS per cache to provide more flexibility.

WisdomPill:
Hello @vedantpuri,

I see your issue, I think it's a valid point so we should have a pool per `cache` entry in settings... would you like to create a PR for this?

It should be quite simple, just pass `CONNECTION_FACTORY` from options and feed it to `pool.get_connection_factory(options=self._options)` [here](https://github.com/jazzband/django-redis/blob/38e630f051d6cc1f54e2e919c5751853282657ee/django_redis/client/default.py#L62) as path and the trick is done.

If you do it yourself I would be glad to review and release a new version pretty quickly.

vedantpuri:
Sure! Here is a proposal:
https://github.com/jazzband/django-redis/pull/620

RealFatCat:
@WisdomPill hi, can you look at https://github.com/jazzband/django-redis/pull/680 please?

RealFatCat:
@WisdomPill can we get a release with this feature?

WisdomPill:
I released it yesterday, thanks @RealFatCat for the PR! 😄

PR #620, opened 2022-10-21
Push Connection Factory setting to options

vedantpuri:
Proposal for #619

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -59,8 +59,12 @@ class DefaultClient:
         self._serializer = serializer_cls(options=self._options)
         self._compressor = compressor_cls(options=self._options)
 
+        # First check if local override provided, else fall back to settings
         connection_factory_path = self._options.get(
-            "CONNECTION_FACTORY", "django_redis.pool.ConnectionFactory"
+            "CONNECTION_FACTORY",
+            settings.get(
+                "DJANGO_REDIS_CONNECTION_FACTORY", "django_redis.pool.ConnectionFactory"
+            ),
         )
         self.connection_factory = pool.get_connection_factory(
             path=connection_factory_path, options=self._options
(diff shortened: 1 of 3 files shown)


django_redis/client/default.py
```diff
@@ -59,7 +59,16 @@ def __init__(self, server, params: Dict[str, Any], backend: BaseCache) -> None:
...
         self._compressor = compressor_cls(options=self._options)
 
-        self.connection_factory = pool.get_connection_factory(options=self._options)
+        # First check if local override provided, else fall back to settings
+        connection_factory_path = self._options.get(
+            "CONNECTION_FACTORY",
+            settings.get(
+                "DJANGO_REDIS_CONNECTION_FACTORY", "django_redis.pool.ConnectionFactory"
+            ),
+        )
```
vedantpuri, line 68: For backwards compatibility

WisdomPill: there is not only 'defaultClient'

WisdomPill (changes requested):
There is not only `DefaultClient` and testing is needed

django_redis/client/default.py
vedantpuri: I didn't follow this comment, could you elaborate >

vedantpuri: Following up on this if you could clarify a bit further

WisdomPill: sorry for the late reply, I mean that there is not only `DefaultClient` but also others... but now I see that others all use connection factory of the parent class which is `DefaultClient`

WisdomPill: Also why moving this logic to `DefaultClient`? it makes testing more difficult... please move it back, add tests and changeling file

terencehonles:
This was implemented in https://github.com/jazzband/django-redis/pull/680

PR #626, opened 2022-11-05
Fix mypy errors

foarsitter:
The github actions fail because of some inconvenient mypy cases, this pull request address them.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -212,7 +212,7 @@ class TestDjangoRedisCache:
 
         if isinstance(cache.client, herd.HerdClient):
             default_timeout = cache.client._backend.default_timeout
-            herd_timeout = (default_timeout + herd.CACHE_HERD_TIMEOUT) * 1000
+            herd_timeout = (default_timeout + herd.CACHE_HERD_TIMEOUT) * 1000  # type: ignore # noqa
             herd_pack_value = cache.client._pack(value, default_timeout)
             mocked_set.assert_called_once_with(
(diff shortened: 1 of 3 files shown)


tests/test_backend.py
```diff
@@ -211,7 +211,7 @@ def test_set_call_empty_pipeline(self, cache: RedisCache, mocker: MockerFixture)
         cache.set(key, value, client=pipeline)
 
         if isinstance(cache.client, herd.HerdClient):
-            default_timeout = cache.client._backend.default_timeout
+            default_timeout = cache.client._backend.default_timeout or 0
```
WisdomPill, line 214: why this?

foarsitter: According to mypy default_timeout can be None

WisdomPill: I mean, it's a test, which can also be not typed technically, but I'm not a fan of using `or` because I find it not readable, can the default be set in another line with an if or something?

foarsitter: Ignoring mypy for that line would be an option with `# type: ignore`.

I totally agree with you that this makes no sense at all.

WisdomPill: I like the `ignore` solution

foarsitter: Added the ignore, feels as the right thing to do over here.

WisdomPill:
thanks for the contribution!

foarsitter:
Thank you for your quick response and maintaining this package!

PR #639, opened 2023-01-06
Fix #638 Lazy access to Django settings.

### LINKED ISSUE #638: Import to django_redis.cache fails
**Describe the bug**
I can't import `django_redis.cache.RedisCache` to extend it, because the module access `django.conf.settings` on import time.

**To Reproduce**

```
$ python -c "import django_redis.cache"
  File "<string>", line 1, in <module>
  File "./lib/python3.11/site-packages/django_redis/cache.py", line 12, in <module>
    DJANGO_REDIS_SCAN_ITERSIZE = getattr(settings, "DJANGO_REDIS_SCAN_ITERSIZE", 10)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./lib/python3.11/site-packages/django/conf/__init__.py", line 87, in __getattr__
    self._setup(name)
  File "./lib/python3.11/site-packages/django/conf/__init__.py", line 67, in _setup
    raise ImproperlyConfigured(
django.core.exceptions.ImproperlyConfigured: Requested setting DJANGO_REDIS_SCAN_ITERSIZE, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```

iurisilvio:


iurisilvio:
Broken tests are not related to another issue https://github.com/jazzband/django-redis/issues/645.

iurisilvio:
HerdClient has no tests, I'm not happy about it, I'll try to write some basic tests at least.

django_redis/cache.py
```diff
@@ -105,7 +103,8 @@ def delete(self, *args, **kwargs):
 
     @omit_exception
     def delete_pattern(self, *args, **kwargs):
-        kwargs["itersize"] = kwargs.get("itersize", DJANGO_REDIS_SCAN_ITERSIZE)
+        if not kwargs.get("itersize"):
```
WisdomPill, line 106: I think it would be cleaner to get eternize from settings in the __init__ method like you did in HerdClient

setup.cfg
```diff
@@ -100,6 +100,7 @@ REDIS =
 passenv = CI GITHUB*
 commands =
   {envpython} -m pytest --cov-report= --ds=settings.sqlite {posargs}
+  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_herd {posargs}
```
WisdomPill, line 103: good catch!

changelog.d/638.removal
WisdomPill, line 1: maybe this file should be a bug fix instead?

tests/test_backend.py
```diff
@@ -495,11 +494,16 @@ def test_delete_pattern_with_custom_count(self, client_mock, cache: RedisCache):
 
     @patch("django_redis.cache.RedisCache.client")
     def test_delete_pattern_with_settings_default_scan_count(
-        self, client_mock, cache: RedisCache
+        self,
+        client_mock,
+        cache: RedisCache,
+        settings: SettingsWrapper,
     ):
+        settings.DJANGO_REDIS_SCAN_ITERSIZE = 30
```
WisdomPill, line 502: can you use `@override_settings()` instead?

tests/conftest.py
```diff
@@ -2,10 +2,16 @@
 
 import pytest
 from django.core.cache import cache as default_cache
+from pytest_django.fixtures import SettingsWrapper
 
 from django_redis.cache import BaseCache
 
 
+@pytest.fixture(autouse=True)
+def patch_settings(settings: SettingsWrapper):
```
WisdomPill, line 11: not sure if this is needed everywhere

WisdomPill (changes requested):
A few comments

tests/test_backend.py
iurisilvio: Done!

changelog.d/638.removal
iurisilvio: Done!

tests/conftest.py
WisdomPill: this is probably not needed, I suggest to try to add this default setting in [here](https://github.com/jazzband/django-redis/blob/master/tests/settings/sqlite_herd.py)

django_redis/cache.py
iurisilvio: Done! Testing become trickier because Django cache reuse the same instance for all tests, so I have to test with a fixture to destroy the instance, change setting and destroy again after test to avoid other tests using a wrong setting somewhere else.

tests/conftest.py
iurisilvio: Done! Moved to `settings_herd.py`.

iurisilvio:
I agreed and did all requested changes, ready for a new review.

WisdomPill (approved):

PR #644, opened 2023-01-07
Fix CI, drop Python 3.6 support.

iurisilvio:


WisdomPill:
why drop the python3.6 support?

iurisilvio:
One good reason to drop 3.6 support is because it is already dead. https://endoflife.date/python

Most libs already dropped this support, like tox. We have to use an old tox lib (`tox<4`) to support 3.6. Other libs will drop support soon.

One reason to not drop it is because Django 3.2 (LTS) still support it. https://endoflife.date/django

iurisilvio:
Here the Python versions Django support. https://docs.djangoproject.com/en/4.1/faq/install/#what-python-version-can-i-use-with-django

iurisilvio:
BUT I agree it should be another issue. For now, I'm happy fixing CI.

Issue #646, opened 2023-01-19
PickleSerializer removed from Django main


iurisilvio:
PickleSerializer was deprecated and removed from Django.

https://code.djangoproject.com/ticket/29708

Actions for Django main are broken. https://github.com/jazzband/django-redis/actions/runs/3956633560/jobs/6776067867#step:7:154

rootart:
@iurisilvio  prepared PR to skip the test that uses PickleSerializer for Django > 4.2. As the `django.contrib.sessions.serializers.PickleSerializer` will be removed entirely it looks like we don't have to perform any other deprecation changes and only keep compatibility with the older Django versions. Let me know what you think about it.

WisdomPill:
but what will be used as default then?

iurisilvio:
Django removed the serializer, but their redis serializer still use pickle.

https://github.com/django/django/commit/c6cb5a0277fce1b87a4b417002289c31f0ee44bc

iurisilvio:
We use django PickleSerializer only to test sessions.

https://docs.djangoproject.com/en/4.1/topics/http/sessions/#session-serialization

Django changed sessions to work with a json serializer and it is their default now. I'm not sure if in the past the default was pickle.

https://github.com/jazzband/django-redis/blob/5bf67ab4408c837e375b3d911738f88a946c0a25/tests/test_session.py#L306

Looks like our test is based on Django test for it and they changed it some time ago. https://github.com/django/django/commit/436862787cbdbd68b0ba20ed8c23b295e3679df3

I think we should just update the test.

iurisilvio:
Pickle was necessary to serialize session expiration, but now Django save it as string.

WisdomPill:
Fixed in #647

PR #647, opened 2023-01-20
#646, fix skip tests for the django.contrib.sessions.serializers.Pick…

rootart:
…leSerializer for django > 4.2


WisdomPill (changes requested):
Thanks for the contribution!

we need a changelog file and fix flake8 first, but now I am thinking which serialised should be used as default

PR #648, opened 2023-01-21
Dropped support for django 2.2 and 3.1

WisdomPill:


WisdomPill:
Failing tests will be fixed in #647

PR #655, opened 2023-02-17
add compressed serializers

KommuSoft:
Often JSON/msgpack/pickle can produce large blobs with a lot of duplicated data.

By using a fast compressor, this can result in both faster writing and reading because the "bandwidth" between the Django layer and the database is smaller, and therefore more efficient. Yes, compressing takes some time as well, but for certain types of data (like pickled dataframes for example), this might pay off, and overall speed up processing.

### DIFF
--- a/django_redis/serializers/compressed.py
+++ b/django_redis/serializers/compressed.py
@@ -1,0 +1,11 @@
+from django.utils.encoding import force_bytes
+
+import lzma
+
+
+class CompressedMixin:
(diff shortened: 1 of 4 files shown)


WisdomPill (commented):
hello @KommuSoft, thanks for the contribution. I wanted to ask you if you were aware that `django_redis` already has compressors? because they already do what you proposed and are more customizable

PR #657, opened 2023-03-02
feat: Allow cache OPTIONS to specify USERNAME

### LINKED ISSUE #656: Pass `USERNAME` to Redis `connection_kwargs`
**Problem Statement**
The username defined in the cache options is not passed to the connection kwargs.

I'm connecting to a Sentinel cluster with the following cache parameters:

```python
CACHES = {
    'default': {
        'BACKEND': 'django_redis.cache.RedisCache',
        'LOCATION': 'redis://redis6.local/0',
        'OPTIONS': {
            'CLIENT_CLASS': 'django_redis.client.SentinelClient',
            'USERNAME': 'my-redis-user',
            'PASSWORD': 'my-very-long-redis-password',
            SENTINEL_KWARGS: {
                'username': 'my-sentinel-user',
                'password': 'my-very-long-sentinel-password',
            },
            'SENTINELS': [
                ('redis6a.local', 26379), ('redis6b.local', 26379), ('redis6c.local', 26379)
            ],
        }
    }
}
```

So the [`SentinelConnectionFactory`](https://github.com/jazzband/django-redis/blob/9d572cd460e1acc0458dfa32a5137eb55b844c52/django_redis/pool.py#L134) is used.

The `SENTINEL_KWARGS` are passed to the [`Sentinel`](https://github.com/jazzband/django-redis/blob/9d572cd460e1acc0458dfa32a5137eb55b844c52/django_redis/pool.py#L153) objects, but the `USERNAME` is not passed to the [`connection_kwargs`](https://github.com/jazzband/django-redis/blob/9d572cd460e1acc0458dfa32a5137eb55b844c52/django_redis/pool.py#L156) param.

**Describe the solution you'd like**
The [`make_connection_params()`](https://github.com/jazzband/django-redis/blob/9d572cd460e1acc0458dfa32a5137eb55b844c52/django_redis/pool.py#L35) method should get the `USERNAME` from the `options`.

```python
def make_connection_params(self, url):

    kwargs = {
        "url": url,
        "parser_class": self.get_parser_cls(),
    }

    # ADD THIS
    username = self.options.get("USERNAME", None)
    if username:
        kwargs["username"] = username

    password = self.options.get("PASSWORD", None)
    if password:
        kwargs["password"] = password

    # [...]

    return kwargs
```

I'm not sure if there are other cases where the `USERNAME` should be passed to the `Sentinel` or `Redis` objects. Maybe in the `ConnectionFactory`'s [`get_connection_pool()`](https://github.com/jazzband/django-redis/blob/9d572cd460e1acc0458dfa32a5137eb55b844c52/django_redis/pool.py#L127) method?

**Describe alternatives you've considered**
-

**Additional context**
-

joachimBurket:
resolves #656


README.rst
```diff
@@ -106,18 +106,20 @@ There are several ways to specify a database number:
...
+username and password in the Cache ``OPTIONS``, or in the URL.
+
 The login for the user ``django`` would look like this:
 
 .. code-block:: python
 
     CACHES = {
         "default": {
             "BACKEND": "django_redis.cache.RedisCache",
-            "LOCATION": "redis://django@localhost:6379/0",
```
WisdomPill, line 118: I would keep both examples if they work

WisdomPill (changes requested):
thanks for the contribution! it would be nice to have a test for this and a change log file

README.rst
joachimBurket: I merged the example in which the username is passed in the URL and the password in the OPTIONS with the example of a URL unsafe password, because I thought it was redundant. Tell me if it suits you as done in [039432c](https://github.com/jazzband/django-redis/pull/657/commits/039432cd21748aa79dcc32541b657c289f74c306)

joachimBurket:
I added the changelog and a test. I wasn't sure where to test the connection params so I renamed the `test_connection_string.py` file in `test_connection_params.py` and added it there with the test of the URL.

WisdomPill (approved):

WisdomPill:
I would love to have tests to connect to a server with credentials

joachimBurket:
Yeaaaah it's been a while sorry :sweat_smile:
@WisdomPill I'm not sure how the missing tests should be implemented, could you add some precision on  this?

terencehonles:
> I would love to have tests to connect to a server with credentials

This would require we have a test setup with creds, but since we're just calling `ConnectionPool.from_url` and merging it with the ``OPTIONS`` this seems like a somewhat low value test. I'm adding a test that just ensures the expected kwargs are created and this is mostly the same as the tests you added except it will include resolving the precedence between the two configuration sources.

terencehonles (approved):
LGTM, thanks

PR #677, opened 2023-08-18
Readme patch to resolve #676

### LINKED ISSUE #676: Automatic upgrade to redis-py version 5.0 causes ImportError if HiredisParser enable
Since v5.0 of the _redis-py_ package was released I'm getting the following error with a new install of my Django project:

`Module "redis.connection" does not define a "HiredisParser" attribute/class`

In my Django project's _requirements.txt_ I have an entry for _django-redis==5.3.0_, but do not pin a version for _redis-py_. I did it that way to always get the latest version that `django-redis` recommends.

**Work around**
I can avoid the error by either:
1. Explicitly pinning _redis-py_ to version 4.6.0
2. Removing support for Hiredis, ie. removing the config line `'PARSER_CLASS': 'redis.connection.HiredisParser'`

jacklinke:
Added note about configuring pluggable parser with hiredis for those who are using redis-py 5, resolving https://github.com/jazzband/django-redis/issues/676

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -754,6 +754,8 @@ used with django-redis.
         "PARSER_CLASS": "redis.connection.HiredisParser",
     }
 
+Note: if using version 5 of redis-py, you should use ``"redis.connection._HiredisParser"`` do to an internal rename of classes within that package.
+
 Pluggable clients
 ~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


WisdomPill (approved):
LGTM!

would you like to create PR for making this automatic? based on the version of `redis-py`

PR #680, opened 2023-09-12
Connection factory goes to cache options

RealFatCat:
Another view for https://github.com/jazzband/django-redis/issues/619

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -736,6 +736,35 @@ In order to enable this functionality you should add the following:
         },
     }
 
+It is also possible to set some caches as sentinels and some as not:
+
+.. code-block:: python
+
+    SENTINELS = [
+        ('sentinel-1', 26379),
+        ('sentinel-2', 26379),
+        ('sentinel-3', 26379),
+    ]
+    CACHES = {
+        "sentinel": {
+            "BACKEND": "django_redis.cache.RedisCache",
+            "LOCATION": "redis://service_name/db",
+            "OPTIONS": {
+                "CLIENT_CLASS": "django_redis.client.SentinelClient",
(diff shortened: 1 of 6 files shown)


tests/test_connection_string.py
```diff
@@ -17,3 +17,24 @@ def test_connection_strings(connection_string: str):
...
+def test_connection_strings_with_conn_factory_in_opts(connection_string: str):
+    sentinel_conn_factory = "django_redis.pool.SentinelConnectionFactory"
+    cf = pool.get_connection_factory(
+        path=None,
+        options={
+            "CONNECTION_FACTORY": sentinel_conn_factory,
+            "SENTINELS": [("127.0.0.1", "26739")],
+        },
+    )
+    res = cf.make_connection_params(connection_string)
```
WisdomPill, line 39: why check connection params?

why not check if CF is an instance of `SentinelConnectionFactory`? I would even use multiple connection factories from `parametrize` and check that the same instance is created

RealFatCat:
Added some tests for `get_connection_factory`.

akostrikov:
Wow, nice feature! I really want to see that in production

WisdomPill (approved):
LGTM!

WisdomPill:
last but not least is to add a changeling file so that when the release is made the `CHANGELOG` is automatic

RealFatCat:
> last but not least is to add a changelog file so that when the release is made the `CHANGELOG` is automatic

done

akostrikov:
@WisdomPill Hello, Anas! Any info on when the next release is going to be?

WisdomPill:
this weekend

Issue #690, opened 2023-11-01
Pytest 7.0 compatibility 


foarsitter:
Due to the release of `pytest-django`, pytest >= 7.0 is now [required](https://github.com/pytest-dev/pytest-django/commit/c3434f65bd4858d94d570345c1f49773597a860b). This is breaking the pipeline.

WisdomPill:
I think we can close this now

PR #691, opened 2023-11-01
Pin pytest to <7

foarsitter:
This PR pins pytest to < 7.0 to mitigate #690


WisdomPill (approved):
LGTM!

you got me! I was just about opening a PR myself! 😄

foarsitter:
@WisdomPill dammit, that would have saved me a lot of trouble! 

As always, thanks for your quick response!

PR #697, opened 2023-11-02
Support pytest7

WisdomPill:
Change setup.cfg in order to support pytest 7


foarsitter:
Didn't expect it to be thats easy. 

Dropping Python 3.6 & 3.7 is inevitable I think.

Issue #598, opened 2022-04-03
Add support for hashmaps


WisdomPill:
**Problem Statement**
`django-redis` should introduce hasmaps and support basic operations, `hset`, `hdel`, `hlen`, `hkeys` and `hexists`.

**Describe the solution you'd like**
Add an interface similar to `redis-py` to support hasps

YashKumarVerma:
@WisdomPill is this one resolved or are we still looking for help here? I'm interested if its needed

WisdomPill:
still looking for an implementation, #654 is a very good attempt at it on the sets side, if you want to propose a solution you're very welcome to do so 😄

LucianaAG:
is available to work on this issue?

WisdomPill:
Hello @LucianaAG I think @kysre is doing all right here, you can try to do something similar to his PR for sets maybe?

WisdomPill:
support for `hset`, `hlen`, `hdel`, `hkeys` and `hexists` has been added thanks to the efforts of @kysre.

Things left to add to make it complete are.

- [ ] tests for in `ShardClient` (this should be fairly easy, replacing `get_client` with method with make_key and get_server based on the key should do the trick like it is done in other `ShardedClient` methods)
- [x] `HDEL`
- [x] `HEXISTS`
- [ ] `HGET` quite important
- [ ] `HGETALL` quite important
- [ ] `HINCRBY`
- [ ] `HINCRYBYFLOAT`
- [x] `HKEYS`
- [x] `HLEN`
- [ ] `HMGET`
- [ ] `HMSET`
- [ ] `HRANDFIELD` not relevant
- [ ] `HSCAN`
- [x] `HSET`
- [ ] `HSETNX` relevant could be implemented in `.hset()`
- [ ] `HSTRLEN` not relevant
- [ ] `HVALS` not relevant


any contribution is welcome, I consider this a beginner friendly issue

rootart:
Hi, @WisdomPill I extended support for the missing hashmap functions and would appreciate your feedback when you have some spare time. 
`HINCRYBYFLOAT` is one of the functions that is missing and I think it would require a bit more work to implement given how we do the decoding/encoding of the values https://github.com/jazzband/django-redis/blob/master/django_redis/client/default.py#L509-L518 I worry if we explicitly store the float value it would come up as incompatible change. Let me know what are your thoughts on this and thanks again.

PR #719, opened 2024-01-26
Add support for hashmaps

kysre:
#598

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -813,3 +813,79 @@ class DefaultClient:
         # Convert to milliseconds
         timeout = int(timeout * 1000)
         return bool(client.pexpire(key, timeout))
+
+    def hset(
+        self,
+        name: str,
+        key: KeyT,
+        value: EncodableT,
+        version: Optional[int] = None,
+        client: Optional[Redis] = None,
+    ) -> int:
+        """
+        Set the value of hash name at key to value.
+        Returns the number of fields added to the hash.
+        """
+        if client is None:
+            client = self.get_client(write=True)
+        nkey = self.make_key(key, version=version)
+        nvalue = self.encode(value)
+        return int(client.hset(name, nkey, nvalue))
+
+    def hdel(
+        self,
+        name: str,
+        key: KeyT,
+        version: Optional[int] = None,
+        client: Optional[Redis] = None,
+    ) -> int:
+        """
+        Remove keys from hash name.
+        Returns the number of fields deleted from the hash.
+        """
+        if client is None:
+            client = self.get_client(write=True)
+        nkey = self.make_key(key, version=version)
+        return int(client.hdel(name, nkey))
+
+    def hlen(
+        self,
+        name: str,
+        client: Optional[Redis] = None,
+    ) -> int:
+        """
+        Return the number of items in hash name.
+        """
+        if client is None:
+            client = self.get_client(write=False)
+        return int(client.hlen(name))
+
+    def hkeys(
+        self,
+        name: str,
+        client: Optional[Redis] = None,
+    ) -> List:
+        """
+        Return a list of keys in hash name.
+        """
+        if client is None:
+            client = self.get_client(write=False)
+        try:
+            return [self.reverse_key(k.decode()) for k in client.hkeys(name)]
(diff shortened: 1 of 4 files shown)


WisdomPill:
thanks for the effort @kysre 

can you merge master on your branch? just so tests might start to run and your PR will look cleaner?

WisdomPill:
never mind I've fixed it

kysre:
Do you suggest that I encode my keys, values like they are encoded in set() function?
```
nkey = self.make_key(key, version=version)
nvalue = self.encode(value)
```

kysre:
I fixed the problem with encoding and decoding values. Now I am getting errors like below in tests:
```
________________________ TestDjangoRedisCache.test_hlen ________________________
  
  self = <tests.test_backend.TestDjangoRedisCache object at 0x7f4c20da6af0>
  cache = <django.utils.connection.ConnectionProxy object at 0x7f4c218c93a0>
  
      def test_hlen(self, cache: RedisCache):
  >       assert cache.hlen("foo_hash3") == 0
  
  tests/test_backend.py:819: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  django_redis/cache.py:29: in _decorator
  django_redis/cache.py:198: in hlen
      return self.client.hlen(*args, **kwargs)
  django_redis/client/default.py:860: in hlen
      client = self.get_client(write=False)
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  self = <django_redis.client.sharded.ShardClient object at 0x7f4c209f5a60>
  args = (), kwargs = {'write': False}
  
      def get_client(self, *args, **kwargs):
  >       raise NotImplementedError
  E       NotImplementedError
  django_redis/client/sharded.py:27: NotImplementedError
  ```
  Should I implement hashmap methods in `django_redis/client/sharded.py` too?

File: django_redis/client/default.py [hunk 1]
```diff
@@ -813,3 +813,79 @@ def touch(
...
+        self,
+        name: str,
+        client: Optional[Redis] = None,
+    ) -> List:
+        """
+        Return a list of keys in hash name.
+        """
+        if client is None:
+            client = self.get_client(write=False)
+        try:
```
WisdomPill, line 873: why the try/catch?

django_redis/client/default.py
```diff
@@ -813,3 +813,79 @@ def touch(
...
+        """
+        if client is None:
+            client = self.get_client(write=False)
+        return int(client.hlen(name))
+
+    def hkeys(
+        self,
+        name: str,
+        client: Optional[Redis] = None,
+    ) -> List:
```
WisdomPill, line 867: list of what?

WisdomPill (changes requested):
Overall looks good, just a couple of comments.

I think the problem with the tests you wrote is the `ShardedClient`, you can decide to skip those tests or implement them in the `ShardedClient`.

If you want your tests to work with shared client just change the `get_client` method with `make_key` and `get_server` based on the key

django_redis/client/default.py
kysre: I implemented hkeys() similar to [keys()](https://github.com/jazzband/django-redis/blob/9c8cf8f32b18bf57bd684ee305e9ba6198e4ef1e/django_redis/client/default.py#L726) function. If you think that the try/catch is redundant, I can remove it.

django_redis/client/default.py
kysre: Fixed it.

kysre:
I decided to skip `ShardedClient` tests for now. If you think it's better to implement them in `ShradedClient`, tell me and I'll do it.

django_redis/client/default.py
WisdomPill: it's okay then, it was implemented way before I took this project, maybe we should discuss it in a separate time

WisdomPill:
it's okay, I would merge it and then add TODOs in the issue so that other people can take it a step further and add support other commands methods too

PR #721, opened 2024-01-30
Add missing methods

LucianaAG:
Continuation of  https://github.com/jazzband/django-redis/pull/654 PR. The previous left it halfway.

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -1022,13 +1022,13 @@ class DefaultClient:
         count: Optional[int] = None,
         version: Optional[int] = None,
         client: Optional[Redis] = None,
-    ) -> Union[set, Any]:
+    ) -> Union[Set, Any]:
         if client is None:
             client = self.get_client(write=True)
 
         key = self.make_key(key, version=version)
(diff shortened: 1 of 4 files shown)


WisdomPill:
@LucianaAG you need to use `Set` from `typing` module because we still support older versions of python.

About changeling you need to create a file inside `changelog.d` with hit name `721.feature` and as content a description of what this PR does, in this way the CHANGELOG file will be updated accordingly when we release and the release notes will have that info as well

django_redis/client/default.py
```diff
@@ -889,3 +889,265 @@ def hexists(
...
+
+    def sscan(
+        self,
+        key: Any,
+        cursor: int = 0,
+        match: Optional[str] = None,
+        count: Optional[int] = None,
+        version: Optional[int] = None,
+        client: Optional[Redis] = None,
+    ) -> set[Any]:
```
WisdomPill, line 1138: you need to change this one to `Set`

WisdomPill (changes requested):
I've pointed out where changes are needed to fix errors in CI

PR #726, opened 2024-02-16
Fix:Expire method fails when using DEFAULT_TIMEOUT

### LINKED ISSUE #724: `Expire` method fails when using `DEFAULT_TIMEOUT`
**Describe the bug**
When attempting to reset the TTL of a key, calling `cache.expire(key, DEFAULT_TIMEOUT)` fails, because the stubbed `DEFAULT_TIMEOUT = object()` is passed through to redis-py rather than an integer.

This does not affect the `touch` method, because that one explicitly updates the timeout with:
```python
if timeout is DEFAULT_TIMEOUT:
    timeout = self._backend.default_timeout
```

**To Reproduce**
1. Run `cache.expire(key, DEFAULT_TIMEOUT)`

**Expected behavior**
The default timeout should be replaced by the backend _integer_ default timeout before sending the value to redis-py

**Stack trace**
```python
  File "my_project/utils/cache.py", line 55, in get_cache
    django_cache.expire(progress_key, DEFAULT_TIMEOUT)
  File "venv/lib/site-packages/django_redis/cache.py", line 29, in _decorator
  File "venv/lib/site-packages/django_redis/cache.py", line 162, in expire
    return self.client.expire(*args, **kwargs)
  File "venv/lib/site-packages/django_redis/client/default.py", line 289, in expire
    return client.expire(key, timeout)
  File "venv/lib/site-packages/redis/client.py", line 1591, in expire
    return self.execute_command('EXPIRE', name, time)
  File "venv/lib/site-packages/redis/client.py", line 900, in execute_command
    conn.send_command(*args)
  File "venv/lib/site-packages/redis/connection.py", line 725, in send_command
    self.send_packed_command(self.pack_command(*args),
  File "venv/lib/site-packages/redis/connection.py", line 775, in pack_command
    for arg in imap(self.encoder.encode, args):
  File "venv/lib/site-packages/redis/connection.py", line 119, in encode
    raise DataError("Invalid input of type: '%s'. Convert to a "
redis.exceptions.DataError: Invalid input of type: 'object'. Convert to a bytes, string, int or float first.
```

 - Python version: 3.9
 - Django Redis Version: 5.4.0
 - Django Version: 3.2

sauravsharma1998:
closes #724


WisdomPill (changes requested):
we need a test and changelog

sauravsharma1998:
@WisdomPill Whenever it's convenient for you, could you please take a moment to review and approve the workflow? Your approval would be greatly appreciated. Thank you!

changelog.d/724.bugfix
```diff
@@ -0,0 +1,3 @@
+Fix for Issue 724(expire method fails when using DEFAULT_TIMEOUT)
+
+https://github.com/jazzband/django-redis/issues/724
```
WisdomPill, line 3: no need to point out the issue just write an explanation

django_redis/client/default.py
```diff
@@ -299,6 +299,15 @@ def expire(
         version: Optional[int] = None,
         client: Optional[Redis] = None,
     ) -> bool:
+        # timeout could be of type int|timedelta
+        # if timeout is DEFAULT_TIMEOUT or None then
+        # use self._backend.default_timeout(django default cache timeout)
+
+        # for some strange reason mypy complains,
+        # saying that timeout type is float | timedelta
```
WisdomPill, line 307: please remove this comment

tests/test_backend.py
```diff
@@ -606,6 +607,10 @@ def test_expire(self, cache: RedisCache):
         assert pytest.approx(ttl) == 20
         assert cache.expire("not-existent-key", 20) is False
 
+    def test_expire_with_default_timeout(self, cache: RedisCache):
+        cache.set("foo", "bar", timeout=None)
+        assert cache.expire("foo", DEFAULT_TIMEOUT) is True
```
WisdomPill, line 612: can you try to expire a non existing key as well?

WisdomPill (changes requested):
we're getting there... could you do the same for `pexpire` as well?

sauravsharma1998:
@WisdomPill Your valuable input has been carefully integrated. Whenever you have a moment, could you kindly review and approve the workflow? Your cooperation is sincerely appreciated. Thank you!

django_redis/client/default.py
```diff
@@ -299,6 +299,9 @@ def expire(
         version: Optional[int] = None,
         client: Optional[Redis] = None,
     ) -> bool:
+        if (timeout is DEFAULT_TIMEOUT) or (timeout is None):
```
WisdomPill, line 302: I would let people set None as timeout and get the exception from the client so that they notice they made a mistake instead of getting a seemingly random expiry time

django_redis/client/default.py
```diff
@@ -315,6 +318,9 @@ def pexpire(
...
```
WisdomPill, line 321: I would let people set None as timeout and get the exception from the client so that they notice they made a mistake instead of getting a seemingly random expiry time

WisdomPill (approved):
LGTM!

just one small thing to remove and we're good

django_redis/client/default.py
sauravsharma1998: Thanks for your feedback @WisdomPill!
Changes are done. Please approve the workflow.

django_redis/client/default.py
WisdomPill: this one is missing

sauravsharma1998: Sorry this was my bad. Fixed!

PR #735, opened 2024-02-28
Fix: Omit tests directory from code coverage

sauravsharma1998:


setup.cfg
```diff
@@ -80,18 +82,18 @@ REDIS =
...
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_lz4 {posargs}
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_msgpack {posargs}
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_sentinel {posargs}
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_sentinel_opts {posargs}
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_sharding {posargs}
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_usock {posargs}
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_zlib {posargs}
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_zstd {posargs}
-  {envpython} -m pytest --cov-append --cov-report= --ds=settings.sqlite_gzip {posargs}
+  {envpython} -m pytest -vv --ignore=tests/* --cov-report= --ds=settings.sqlite {posargs}
```
WisdomPill, line 85: why do we need these ignores?

setup.cfg
```diff
@@ -115,7 +117,7 @@ envdir={toxworkdir}/lint
 commands =
     black: black --target-version py36 {posargs:--check --diff} setup.py django_redis/ tests/
     ruff: ruff {posargs:check --show-fixes} django_redis/ tests/
-    mypy: mypy {posargs:--cobertura-xml-report .} django_redis tests
+    mypy: mypy {posargs:--cobertura-xml-report .} django_redis
```
WisdomPill, line 120: please restore this, as also tests should be checked for typing errors so that we can notice breaking changes

WisdomPill (changes requested):

WisdomPill:
also changelog is missing

setup.cfg
sauravsharma1998: Sure @WisdomPill ! This was just experimental. Any suggestions to avoid tests directory from coverage report?

sauravsharma1998:
@WisdomPill Added new config file codecov.yml to manage the coverage on the basis of flags and directory.

Motivation - https://docs.codecov.com/docs/github-5b-setting-coverage-standards-with-flags

WisdomPill:
we still need coverage on tests for mypy

codecov.yml
WisdomPill, line 1: why not change the codecov in .github?

sauravsharma1998: @WisdomPill Do we require to upload mypy report on coverage.io?

sauravsharma1998: By default coverage.io will merge all the coverages, that's why coverage percentage is lower than expected.

WisdomPill: we are using codecov, coverage.io does not exist as far as I know. I was asking why did you not change [this file instead](https://github.com/jazzband/django-redis/blob/master/.github/codecov.yml) and created a new one.

And yes we need to upload the coverage of mypy too to codecov

sauravsharma1998: Got your point @WisdomPill !

With coverage.io I meant to https://app.codecov.io/

Could you provide any idea, Why mypy coverage is low? It would be helpful to think further in combing both type of reports.

sauravsharma1998: MyPy's type checking does not fall under the category of code coverage.

Code coverage typically refers to the measurement of how much of your codebase is exercised by your test suite during execution. It's a metric that indicates the percentage of lines, branches, or other code elements that have been executed by your tests. Code coverage tools like coverage.py or JaCoCo track which parts of your code are executed during testing and generate reports to help you understand how thoroughly your tests exercise your code.

On the other hand, MyPy is a static type checker for Python code. It analyzes your codebase to identify potential type-related issues, inconsistencies, or violations of type hints. MyPy operates statically, meaning it analyzes the code without actually running it. Instead of measuring code coverage, MyPy focuses on ensuring type safety and correctness by enforcing type annotations and performing type inference.

@WisdomPill Please correct me if I am wrong.

sauravsharma1998:
@WisdomPill Instead of uploading mypy report to codecov.io we can generated html formatted report of mypy(Which will satisfy purposes for both code coverage and type checking coverage)

Sample Screenshot:
image

If its okay for you, we can proceed. Else have to find another way.

terencehonles:
This doesn't appear to be needed, the coverage reporting does not show the tests when I ran it right now or looking at codecov.io

Issue #597, opened 2022-04-03
Add support for sets


WisdomPill:
**Problem Statement**
`django-redis` should introduce sets and support basic operations, `sadd`, `sremove`, `smembers`, `sismembers` and `spop`.

**Describe the solution you'd like**
Add an interface similar to `redis-py` to support set

vedantpuri:
Any updates on this ?

WisdomPill:
nope, if you like to open a PR I would be happy to review it! 😁

YashKumarVerma:
isn't this already been implemented?

WisdomPill:
no #654 is supposed to implement it, very little is missing, feel free to recycle that work and mention the previous contributor

LucianaAG:
is available to work on this issue?

WisdomPill:
@LucianaAG feel free to open a PR if you want

PR #743, opened 2024-06-10
Add support for the set functions from issue #597

rootart:
@WisdomPill I took over some work that was already done here https://github.com/jazzband/django-redis/pull/654. Will appreciate your support in reviewing it.

### DIFF
--- a/django_redis/client/default.py
+++ b/django_redis/client/default.py
@@ -517,6 +517,17 @@ class DefaultClient:
 
         return value
 
+    def _decode_iterable_result(
+        self, result: Any, covert_to_set: bool = True
+    ) -> Union[List[Any], None, Any]:
+        if result is None:
+            return None
+        if isinstance(result, list):
+            if covert_to_set:
+                return {self.decode(value) for value in result}
+            return [self.decode(value) for value in result]
+        return self.decode(result)
+
     def get_many(
         self,
         keys: Iterable[KeyT],
@@ -828,7 +839,7 @@ class DefaultClient:
         *keys: KeyT,
         version: Optional[int] = None,
         client: Optional[Redis] = None,
-    ) -> Set:
+    ) -> Set[Any]:
         if client is None:
             client = self.get_client(write=False)
 
@@ -855,7 +866,7 @@ class DefaultClient:
         *keys: KeyT,
         version: Optional[int] = None,
         client: Optional[Redis] = None,
-    ) -> Set:
+    ) -> Set[Any]:
         if client is None:
             client = self.get_client(write=False)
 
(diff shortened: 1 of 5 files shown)


django_redis/client/default.py
WisdomPill, line 1: when returning a `Set` can we have the container type as well?

django_redis/client/default.py
```diff
@@ -778,6 +797,265 @@ def make_pattern(
...
+    ) -> int:
+        if client is None:
+            client = self.get_client(write=False)
+
+        key = self.make_key(key, version=version)
+        return int(client.scard(key))
+
+    def sdiff(
+        self,
+        *keys: KeyT,
```
WisdomPill, line 828: do you know what happens in Redis if we give it multiple keys and some are duplicate?

WisdomPill: are you sure the type is correct for `keys`?

django_redis/client/default.py
```diff
@@ -778,6 +797,265 @@ def make_pattern(
...
+        if client is None:
+            client = self.get_client(write=True)
+
+        nkey = self.make_key(key, version=version)
+        result = client.spop(nkey, count)
+        if result is None:
+            return None
+        if isinstance(result, list):
+            return {self.decode(value) for value in result}
+        return self.decode(result)
```
WisdomPill, line 952: could we have an utility method for this?

django_redis/client/sharded.py
```diff
@@ -258,7 +260,8 @@ def incr_version(self, key, delta=1, version=None, client=None):
             raise ConnectionInterrupted(connection=client) from e
 
         if value is None:
-            raise ValueError("Key '%s' not found" % key)
+            err_msg = f"Key '{key}' not found"
```
WisdomPill, line 263: oops, sorry for this, I've already taken care of it in another PR

WisdomPill (approved):
Terrific job!

I just have a couple of questions on typing, but apart from that this PR looks great.

P.S. sorry for stepping on your foot for pre-commit and ruff, I opted to merge the pre-commit autoupdate instead of different PRs

django_redis/client/sharded.py
rootart: No worries, I'm glad that it was fixed.

django_redis/client/default.py
rootart: Sure, I somehow missed that. I don't see a clear way how we can use generics or have better type support rather than `Set[Any]` for the results.  Do you have other suggestions?

django_redis/client/default.py
WisdomPill: today I've learned something, thanks man!

PR #744, opened 2024-06-14
Fix, warnings to prefix select, ignore, flake8-tidy-imports with lint…

rootart:
… in ruff config


WisdomPill:
thank you very much @rootart but I think it's better to do this in #739, I already pushed some fixes

PR #746, opened 2024-06-23
Issue #598 extend support for the hashmap functions

rootart:


tests/test_backend.py
```diff
@@ -810,17 +810,61 @@ def test_clear(self, cache: RedisCache):
...
         cache.hset("foo_hash1", "foo1", "bar1")
         cache.hset("foo_hash1", "foo2", "bar2")
         assert cache.hlen("foo_hash1") == 2
         assert cache.hexists("foo_hash1", "foo1")
         assert cache.hexists("foo_hash1", "foo2")
 
+    def test_hsetnx(self, cache: RedisCache):
+        result_foo1 = cache.hsetnx("foo_hash1", "foo1", "bar1")
+        result_foo2 = cache.hsetnx("foo_hash1", "foo2", "bar2")
+        result_foo2_1 = cache.hsetnx("foo_hash1", "foo2", "bar2")
```
WisdomPill, line 822: could you change the value written?
```suggestion
        result_foo2_1 = cache.hsetnx("foo_hash1", "foo2", "bar3")
```

WisdomPill (changes requested):
looks great, but please add versioning in another PR so that it can be also tested separately

WisdomPill:
@rootart do you have time to finish this PR? otherwise I can take it... that said, thanks for the effort you've put in this project so far, it is highly appreciated 😄

rootart:
I'm happy to come back and continue some work on the PR 😄 
> looks great, but please add versioning in another PR so that it can be also tested separately

Would you suggest deleting versioning, even in the newly added interfaces? There is definitely a dependency on the `hset`, and the logic for the key generation must align with the other methods, like in `hmget`. I updated the existing functions with correct support of versioning in this PR https://github.com/jazzband/django-redis/pull/768 if we decide to start with it first.

2ykwang:
It seems like some work is duplicated; could you please take a look at my PR as well? @rootart @WisdomPill

PR #749, opened 2024-09-06
Fix sentinel mode read-write separation

JackiLin:
When I attempted to set up read-write separation in Sentinel mode by configuring redis://redis-master/db?is_master=0, I noticed that both read and write requests were still being directed to the master node, without any separation. After reviewing and debugging the source code, I found that the issue was due to redis-py being unable to correctly parse the is_master parameter, and there was also an error in how the project handled is_master. The pool.is_master was only set to False after retrieving the redis.sentinel.SentinelConnectionPool object from redis-py, which resulted in the system supposedly using the slave node, but actually using the master node.
image

### DIFF
--- a/django_redis/pool.py
+++ b/django_redis/pool.py
@@ -1,5 +1,5 @@
 from typing import Dict
-from urllib.parse import parse_qs, urlparse,urlencode,urlunparse
+from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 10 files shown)


django_redis/pool.py
```diff
@@ -163,16 +163,27 @@ def get_connection_pool(self, params):
...
+        query_params = parse_qs(url.query)
+        is_master = query_params.get("is_master")
         if is_master:
-            pool.is_master = to_bool(is_master[0])
+            cp_params["is_master"] = to_bool(is_master[0])
+        # then remove the "is_master" query string from the URL
+        # so it doesn't interfere with the SentinelConnectionPool constructor
+        if "is_master" in query_params:
+            del query_params["is_master"]
+        new_query = urlencode(query_params, doseq=True)
```
WisdomPill, line 176: why do you need this if you deleted the key from the original `query_params` variable?

WisdomPill (approved):
LGTM!

django_redis/pool.py
JackiLin: If this parameter is not deleted, it cannot be correctly parsed as bool type in redis-py, but as str type. This will cause it to use the master node for reading and writing even if is_master=0 is set.

WisdomPill: yeah, sure, my bad

WisdomPill: please merge master into your branch and solve conflicts then

WisdomPill (approved):
LGTM!

django_redis/pool.py
JackiLin: I have merged and resolved the conflicts

PR #756, opened 2024-10-26
Added support for Python 3.13

ulgens:


WisdomPill:
`redis-py` still doesn't support python3.13 officially, but we can give it a go.

WisdomPill:
Any plans for it? @Chayim @dvora-h

foarsitter:
@ulgens do not forget to add the classifier for python 3.13 to `setup.cfg`.

ulgens:
@WisdomPill I was just checking `redis-py`, I'll try to create a PR there too, hopefully in the following week.

setup.cfg
```diff
@@ -60,9 +61,10 @@ envlist =
     # tests against released versions
     py{38,39}-dj{42}-redislatest
     py{310,311,312}-dj{42,50,51}-redislatest
+    py313-dj51-redislatest
```
WisdomPill, line 64: why not do like on line 67?

ulgens: This line aims to add Python 3.13 (supported only by Django 5.1) and the latest Redis to the matrix. Not sure how it can be more like line 67, an example would be nice.

ulgens:
> @ulgens do not forget to add the classifier for python 3.13 to `setup.cfg`.

It's in place right now. Thank for the reminder.

setup.cfg
WisdomPill:
```suggestion
    py313-dj51-redis{latest,master}
```
and remove `py313-dj51-redismaster` from line 66

ulgens: Oh, I see. I can apply that, but there is contextual grouping between `tests against released versions` and `tests against unreleased versions` in that block. Do we want to change that?

WisdomPill: `python3.13` is not yet supported by `redis-py` officially so I guess we can consider it unreleased 😄

chayim:
@vladvildanov I think this is you now.

robd003:
@ulgens / @WisdomPill / @vladvildanov Just wanted to nudge and see if this could get moved forward. Redis-py has had Python 3.13 support for a while.

setup.cfg
ulgens: `redis-py` started supporting Python 3.13 in their 6.0.0. release: https://github.com/redis/redis-py/releases/tag/v6.0.0

.github/workflows/ci.yml
```diff
@@ -37,21 +37,26 @@ jobs:
           - django-version: '4.2'
             redis-version: 'latest'
             python-version: '3.9'
+          
+          # Django 5.1 and python 3.13 with latest redis
+          - django-version: '5.1'
```
WisdomPill, line 42: why not django 5.2?

ulgens: Updated.

robd003:
Are we good to go here?

WisdomPill (approved):

PR #757, opened 2024-10-27
Enabled parallel tests

WisdomPill:
Many things have been implemented in order to improve tests

- tests are now generated using `pytest_generate_tests` so now they can be run with a single `pytest` command
- tests are now parallelised using `pytest-xdist`
- settings use different databases now
- `docker-compose` is used to provision the environment for standalone and sentinel setups


ulgens (changes requested):
I feel like the PR is trying to achieve too much at once. Each one of the four items listed in description calls for their own PR. I guess, xdist could be the minimal start point.

WisdomPill:
@ulgens `xdist` cannot work without `pytest_generate_tests`... parallelisation can't work on the same db and there is `pytest-django` that can't work with xdist easily in this case... 😕 
so the only thing that can be separate is probably the `docker-compose` part which is barely a refactoring of what was there before... 

now I see some issues with overriding `settings`, because I removed `pytest-django`, but I think I have an idea to solve them, will update this PR another day when I'll have time and willingness to spend it on this project 😄 .

I didn't find a way to make `pytest-django` work with `pytest_generate_tests` due to the way `pytest-django` initialises django settings, if you have something in mind I could lower the scope of this PR, but imho it does little more than possible to go towards the goal of parallelizing tests and make them faster

I consider going from 5 minutes per env to 1 minute a big improvement tbh so I won't surrender that easily on this PR and try to close it soon 😄

ulgens:
@WisdomPill I remember managing similar needs for different projects with relatively simpler changes but apparently I'm missing a set of details. I'm concerned about the project's history: Having a PR with this size and many changes merged into the repo makes it harder for further contributors to understand.

> I consider going from 5 minutes per env to 1 minute a big improvement tbh so I won't surrender that easily on this PR and try to close it soon 😄

Mad respect there 😄 

>  if you have something in mind I could lower the scope of this PR

I'd like to do some testing with xdist and minimal changes before monday if that's okay. I have no info about the urgency of these changes, but my goal is to be back soon with a recommendation or to discard my review.

WisdomPill:
>  I'm concerned about the project's history: Having a PR with this size and many changes merged into the repo makes it harder for further contributors to understand.

Actually the size of the PR is due to the rewrite of `test_sessions.py` so that it can use a `session` fixture so that settings can be generated with `pytest_generate_tests`, to be fair those tests should be checked against master of django because I believe there are some slight changes.

> I'd like to do some testing with xdist and minimal changes before monday if that's okay. I have no info about the urgency of these changes, but my goal is to be back soon with a recommendation or to discard my review.

please be my guest, I am open to any improvement and I am really thankful for the time and effort you're spending in the project. You're welcome to propose any changes outside of this PR scope as well, your help is deeply appreciated.
Last but not least, if you need more time just write here, I am not in a hurry, even if I fix the issue I can wait, although I have to say merging this improvement is highly tempting 😄 
There are other things to check and manage outside of this PR that are more important for a release.

WisdomPill:
@ulgens I would merge this PR if you don't have any updates

ulgens:
@WisdomPill I wasn't able to allocate the time I was planning, sorry for that. Please go ahead.

PR #765, opened 2025-03-08
fix incorrect namespacing applied to Redis hash field keys

2ykwang:
This changes addresses an issue where namespacing was applied to the field instead of the key.

**behavior:**
- before: Namespacing was applied to the field parameter (e.g, hash key) when it should have been applied to the key parameter.
- after: Namespacing is correctly applied to the key parameter, leaving the field name unaltered.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -113,9 +113,21 @@ lint.select = [
 ]
 
 lint.ignore = [
-  "COM812", # missing trailing comma, covered by black
-  "ANN101", # ignore missing type annotation in self parameter
-  "S311", # ignore Standard pseudo-random generators because they are not used for cryptographic purposes
+  # missing trailing comma, covered by black
+  # docs: https://docs.astral.sh/ruff/rules/missing-trailing-comma/
+  "COM812",
+  # ignore missing type annotation in self parameter
+  # docs: https://docs.astral.sh/ruff/rules/missing-type-self/#missing-type-self-ann101
+  "ANN101",  
+  # ignore builtin import shadowing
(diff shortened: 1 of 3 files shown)


2ykwang:
should I look into something specific about the GitHub CI failure? @WisdomPill

WisdomPill:
hello @2ykwang,

thanks for your contribution! 

so only changelog and ruff are left to fix:
- changelog needs a file called `changelog.d/765.bugfix` because this PR is number 765 and its reason is a bug fix, inside the file write a description of what the PR is fixing, this will later be put into the changelog when releasing automatically with `towncrier`
- as for ruff you can remove `ANN101` and add `A004` [here](https://github.com/jazzband/django-redis/blob/ed279990856373f59dc7d76644d4a02426d4eca8/.ruff.toml#L116) and it will solve itself

2ykwang:
I’ve added the changelog  @WisdomPill

WisdomPill (changes requested):
LGTM, I wonder how tests didn't break. Could you try using different version and checking the key in the tests?

Also please fix ruff's error

WisdomPill:
also FYI, there's no need to mention me in a comment, just ask for a review, I will see the notifications and I will be happy to review 😄

2ykwang:
> LGTM, I wonder how tests didn't break. Could you try using different version and checking the key in the tests?

because `hset`, `hexists`, and all other methods that require keys use the same namespacing logic via make_key, ensuring consistent key generation.
and there are no test cases that specifically verify the namespacing of hash keys.

terencehonles:
This has been merged via #812

PR #768, opened 2025-03-25
Add versioning to the hashmap functions when generating keys

rootart:


terencehonles:
I think the bug was actually #812 by looking at your diff, so please open a new PR if the current code is missing a version somewhere (I only checked methods you changed).

PR #772, opened 2025-05-28
Drop support for Python 3.8 (EOL)

ulgens:


django_redis/client/default.py
```diff
@@ -842,7 +838,7 @@ def sdiff(
         *keys: KeyT,
         version: Optional[int] = None,
         client: Optional[Redis] = None,
-    ) -> Set[Any]:
+    ) -> builtins.set[Any]:
```
WisdomPill, line 841: why not just `set`?

tests/test_cache_options.py
```diff
@@ -1,5 +1,6 @@
 import copy
-from typing import Iterable, cast
+from collections.abc import Iterable
```
WisdomPill, line 2: why Iterable was moved?

WisdomPill (changes requested):
thanks for your contribution, I've left some comments which need to be addressed before going forward

django_redis/client/sharded.py
```diff
@@ -1,7 +1,9 @@
+import builtins
 import re
 from collections import OrderedDict
+from collections.abc import Iterator
 from datetime import datetime
-from typing import Any, Iterator, List, Optional, Set, Union
```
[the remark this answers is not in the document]
ulgens: It's imported via `from collections.abc import Iterator` now, as a result of https://github.com/asottile/reorder-python-imports?tab=readme-ov-file#rewriting-pep-585-typing-imports

django_redis/client/default.py
ulgens: `set` shadows `django_redis.client.default.DefaultClient.set`. We have the same case with all classes that implements a `.set()`method. I'm open for alternative suggestions.

PR #778, opened 2025-06-24
Replace black and ruff calls with pre-commit

ulgens:


WisdomPill (approved):
you know what would also be interesting to do, but it is up to you to agree and do it or not.
Use ruff also for formatting, it is almost compatible with black and in most cases there are no changes

ulgens:
@WisdomPill I was thinking of implementing that in a different PR. One extra thing I have in mind is to wrap mypy with pre-commit and remove it from the test envs list; but I think that will require a small tuning for the coverage upload.

PR #781, opened 2025-06-26
Clean Ruff config

ulgens:
Before enabling ruff format and removing black, I wanted to improve & clean the ruff config.

* Update ruff-check hook id: https://github.com/astral-sh/ruff/issues/10171
* Apply alphabetical ordering to selected ruff rules, to improve readability and maintainability
* Add doc and tool links with the selected ruff rules, to make it easier to understand what those rules do
* Improve the organization of sections in ruff config.

The idea / goal is to provide a cleaner ruff setup which is easier to understand and maintain.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   rev: v0.12.0
   hooks:
-  - id: ruff
+  - id: ruff-check
     args: [ --fix, --exit-non-zero-on-fix , --show-fixes]
 
 - repo: https://github.com/psf/black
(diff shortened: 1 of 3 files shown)


WisdomPill:
can you summarise what you did? and why you've done it?

WisdomPill:
well @ulgens mine was also taken from commercial projects, but in general I am not against your proposal. It's just a matter of taste 😄  it would be nice to see which ones you've added and which ones did you remove, I will take some time to dig on that. In my current job there's another set of rules so again, it's just a matter of subjective taste (when I've added ruff I was working in another place and actually that set of rules was brought from a guy that was working in another place again 😂 ).

ulgens:
>  it would be nice to see which ones you've added and which ones did you remove

Nothing added or removed. I limited this PR to refactoring, to not require that effort to understand what is changed. No change in the config, just alphabetical ordering and extra comments.

PR #796, opened 2025-10-20
feat: update `RedisCache.close` to check for client before closing

wloggains-mmllc:
**Problem**
There are some situations where the `_client` in RedisCache is not instantiated when the `close()` method is called.  We have ran into this specifically with [django.core.cache.CacheHandler.close_caches()](https://github.com/django/django/blob/main/django/core/cache/__init__.py#L61).  While it does not cause an error, it is an inefficiency.

**Solution**
This pull request makes an update to `django_redis/cache.py::RedisCache.close()` to prevent unnecessary initialization of the `_client`.

The close method now checks if `_client is not None` before attempting to close the client.

I have also added a test.


kingbuzzman (approved):

terencehonles (approved):
This seems like a pretty small optimization, but LGTM nevertheless.

Issue #797, opened 2025-11-09
Add support for sorted set ZSET operations


oliverhaas:
**Problem Statement**

Django-redis supports set operations but not sorted sets, forcing you to use the raw Redis client and lose serialization/connection management.

**Describe the solution you'd like**

Add sorted set operations (zadd, zcard, zcount, zincrby, zpopmin/zpopmax, zrange/zrangebyscore, zrank, zrem, zremrangebyscore, zrevrange, zrevrangebyscore, zscore) with proper serialization, zadd flags support, and key versioning.

**Describe alternatives you've considered**

Raw redis-py loses serialization. Custom backend or external package is overkill.

**Additional context**

Have a working mixin implementation. 336 tests, all checks passing. Ready to PR in a second as a PoC. Feedback appreciated.

PR #798, opened 2025-11-09
feat: add sorted set operations and mixins for RedisCache

oliverhaas:
See #797.

I can always clean this up or change the structure, but this is how it made sense to me right now.

I can also add more commands to this PR or in a next PR for sorted sets, but these are the main ones I would say, and I'm trying to not blow this up too much.

If there is nothing majorly wrong with this, I'd like to add like another 2-3 PRs worth of commands soon, since I'm using Redis quite heavily right now and I'd rather not have too many custom code in my projects which should probably just go into django-redis. From my point of view, anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ dump.rdb
 .coverage
 coverage.xml
 cobertura.xml
+CLAUDE.md
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -178,3 +178,6 @@ ban-relative-imports = "all"
 
 # pickle is used on purpose and its use is discouraged
 "django_redis/serializers/pickle.py" = ["S301"]
+
+# min/max are official Redis parameter names matching redis-py API
+"django_redis/client/mixins/sorted_sets.py" = ["A002"]
--- a/changelog.d/797.feature
+++ b/changelog.d/797.feature
@@ -1,0 +1,1 @@
+Add sorted set operations (zadd, zrange, zrem, etc.) and mixins for RedisCache
--- a/django_redis/cache.py
+++ b/django_redis/cache.py
@@ -278,3 +278,60 @@ class RedisCache(BaseCache):
     @omit_exception
     def hexists(self, *args, **kwargs):
         return self.client.hexists(*args, **kwargs)
+
+    # Sorted Set Operations
+    @omit_exception
+    def zadd(self, *args, **kwargs):
+        return self.client.zadd(*args, **kwargs)
+
+    @omit_exception
+    def zcard(self, *args, **kwargs):
+        return self.client.zcard(*args, **kwargs)
+
+    @omit_exception
+    def zcount(self, *args, **kwargs):
+        return self.client.zcount(*args, **kwargs)
+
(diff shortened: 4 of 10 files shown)


WisdomPill:
looks good!

but this change of structure will need to become consistent with the other kind of commands too, are you willing to make PRs to separate the rest of the commands?

I would say, hash and set also need their own mixins...

Last but not least, how do you think could bring benefits for going async?

oliverhaas:
> looks good!
> 
> but this change of structure will need to become consistent with the other kind of commands too, are you willing to make PRs to separate the rest of the commands?
> 
> I would say, hash and set also need their own mixins...
> 
> Last but not least, how do you think could bring benefits for going async?

What I would personally would aim for in the near future is:
- Add most important list commands (as mixin)
- Add more hash commands (+ refactor to mixin)
- Refactor set commands to mixin (don't think anything major is missing here...?)
- Add some more general redis commands like `.type()` and `.info()`. I personally would like to have these, but they are definitely a bit more awkward to have in a more general sense. But as long es these commands will work with redis and valkey in the foreseeable future, it's probably okay. Also, nobody has to use them if they do not want to :smile: 

If that sounds good I'm hoping to continue basically today. Except for the last point there is really not much to it. You could review after you've seen the whole package of those 4-5 PRs, as long as it's realistic from your end these will make it into django-redis without anything big on top of what I've laid out.


Regarding async: I personally don't have much use for async unfortunately, just because my current employer went with sync, although I would really like to use it more. It's not clear to me what the goal and approach would be for django-redis. If you have some github issue or something I'm not seeing let me know.

WisdomPill (approved):

WisdomPill:
thanks @oliverhaas 

the only thing which I don't see any reason to do it is, 

>Add some more general redis commands like .type() and .info(). I personally would like to have these, but they are definitely a bit more awkward to have in a more general sense. But as long es these commands will work with redis and valkey in the foreseeable future, it's probably okay. Also, nobody has to use them if they do not want to 😄

you can always get the direct client connection and run those commands, those are not really Django/python specific... how would you approach the type command? do you want to translate it to a python type? otherwise I don't see any reason to expose all of `redis-py` to `django-redis`

oliverhaas:
> thanks @oliverhaas
> 
> the only thing which I don't see any reason to do it is,
> 
> > Add some more general redis commands like .type() and .info(). I personally would like to have these, but they are definitely a bit more awkward to have in a more general sense. But as long es these commands will work with redis and valkey in the foreseeable future, it's probably okay. Also, nobody has to use them if they do not want to 😄
> 
> you can always get the direct client connection and run those commands, those are not really Django/python specific... how would you approach the type command? do you want to translate it to a python type? otherwise I don't see any reason to expose all of `redis-py` to `django-redis`

I think you're probabky right.
I might need to think about it again, but probably I got just caught up too much in the "I need to decompress and decode everything"-mindset which affects all other commands.

WisdomPill:
@oliverhaas thank you for your work! can't wait to see more of it and I really like the ideas you brought to the project!

PR #803, opened 2025-12-13
refactor: extract set operations into SetMixin

oliverhaas:


terencehonles:
Same as #804 

> @oliverhaas do you mind sync'ing master? You can probably use the `--ours` merge conflict resolve strategy and then run `pre-commit` until it's stable to fix changes due to ruff causing conflicts. That should get you most of the way and then you can look at the diff to see if there are other changes you need to make. Alternatively, and possibly easier, you can just move the methods again.
> 
> Once you have done so, can you add a test that runs a diff against the [redis.commands.core](https://github.com/redis/redis-py/blob/master/redis/commands/core.py) class you are mirroring and ensure that all methods are implemented (possibly with an exclude/skip list for methods that don't make sense here or maybe need to be implemented later if there are many)

Issue #766, opened 2025-03-13
Support redis-py 4.0.2 Hash ops in django-redis


2ykwang:
**Problem Statement**

django-redis doesn’t implement all the hash commands available in redis-py (4.0.2). 
I plan to work on implementing the missing commands, 
I’d appreciate any feedback on this proposal, especially regarding potential edge cases or test strategies.

**Additional context**

**Currently supported hash commands**

| Hash Command    | django-redis | [redis-py 4.0.2](https://github.com/redis/redis-py/blob/v4.0.2/redis/commands/core.py#L3302) | [redis-py 5.2.0](https://github.com/redis/redis-py/blob/v5.2.0/redis/commands/core.py#L4940) |
|-----------------|--------------|----------------|----------------|
| HDEL            | V            | V              | V              |
| HEXISTS         | V            | V              | V              |
| HGET            |              | V              | V              |
| HGETALL         |              | V              | V              |
| HINCRBY         |              | V              | V              |
| HINCRBYFLOAT    |              | V              | V              |
| HKEYS           | V            | V              | V              |
| HLEN            | V            | V              | V              |
| HSET            | V            | V              | V              |
| HSETNX          |              | V              | V              |
| HMSET           |              | V              | V              |
| HMGET           |              | V              | V              |
| HVALS           |              | V              | V              |
| HSTRLEN         |              | V              | V              |
| HEXPIRE         |              |                | V              |
| HPEXPIRE        |              |                | V              |
| HEXPIREAT       |              |                | V              |
| HPEXPIREAT      |              |                | V              |
| HPERSIST        |              |                | V              |
| HEXPIRETIME     |              |                | V              |
| HPEXPIRETIME    |              |                | V              |
| HTTL            |              |                | V              |
| HPTTL           |              |                | V              |

 
related issue:
- #736 
- #741 
- #373

PR #806, opened 2025-12-15
 Fixing issue #766

MehrazRumman:
### Description 
issue #766 
Added full hash command coverage to the cache client `(hget/hgetall/hmget/hmset via mapping/hsetnx/hincrby/hincrbyfloat/hvals/hstrlen)` and exposed them through the cache facade, keeping key/version handling consistent. Tweaked decoding to handle float string responses (needed for hincrbyfloat). Added regression tests for the new hash behaviors (reads, bulk ops, conditional set, numeric increments, value listing, length).


MehrazRumman:
@WisdomPill  Could you please review this PR when you have time? Thank you.

Issue #808, opened 2025-12-17
Why test in CI is failing ?


MehrazRumman:
I ran the test locally everything is fine. But I saw tests are failing. so I started debugging. Is there any clue why some tests are failing ?

MehrazRumman:
I am trying to fix the CI !

terencehonles:
I'm closing this because the CI seems to be green (I also made an update, and confirmed it is green right now)

PR #809, opened 2025-12-17
Fixing failed test

MehrazRumman:
see #808 

Trying to fix the failed test in CI

- [x] Fixing TestSortedSetOperations.test_zscore[sqlite_gzip]
- [x]  Fixing TestDjangoRedisCache.test_set_call_empty_pipeline[sqlite_herd]


WisdomPill:
why so many new methods were added?

tests/test_backend.py
```diff
@@ -22,11 +23,13 @@
 @pytest.fixture
 def patch_itersize_setting() -> Iterable[None]:
     # destroy cache to force recreation with overriden settings
-    del caches["default"]
+    with suppress(AttributeError):
```
WisdomPill, line 26: perhaps this should be solved somehow different? why is it needed? which Django version introduced this error?

django_redis/client/default.py
```diff
@@ -508,6 +508,8 @@ def decode(self, value: EncodableT) -> Any:
         try:
             value = int(value)
         except (ValueError, TypeError):
+            with suppress(ValueError, TypeError):
+                return float(value)
```
WisdomPill, line 512: we do not have incrbyfloat, why add it to hash commands?

WisdomPill:
I would have added separately the changes, this is not just a fix

MehrazRumman:
> why so many new methods were added?

I was trying to solve this : https://github.com/jazzband/django-redis/issues/766

ulgens (changes requested):
Not sure why my review was requested, but the PR seems to be handling multiple things context-wise. I wouldn't merge this without a clear focus first.

terencehonles:
@MehrazRumman it looks like you merged the commits for your fix (gzip related stuff) with #806 , can you make sure everything is properly in #806 that should be in #806 and remove those changes from this branch. I suspect you may just want to cherry pick https://github.com/jazzband/django-redis/pull/809/commits/190668e6fa2302287ede00a521e99e4a44cdd87a, https://github.com/jazzband/django-redis/pull/809/commits/5e4ed9d6e3e06d947c0114aa633eb526ac6cf052, and https://github.com/jazzband/django-redis/pull/809/commits/01c59bcd967bca692416e7018bcab2af8e3163e7 and re-push that to this branch.

It looks like you have some other code churn, but you should figure out if they are related or even intended, and if you want open another PR that just makes those changes.

PR #811, opened 2025-12-24
refactor: rename 'name' parameter to 'key' in sorted set methods

oliverhaas:
This is a follow up of https://github.com/jazzband/django-redis/pull/798.

I've just chosen a bad arg naming, which I only noticed when I worked on the hash methods.

**NOTE:** redis-py unfortunately was the cause for this confusion. What the Redis docs call `key` is `name` in the redis-py code. I suggest that this project stays with `key` for well... the key, just like apparently everyone (incl. Redis and Valkey docs) does except the redis-py code.


terencehonles (approved):
Thanks, LGTM

PR #812, opened 2025-12-24
fix: hash method parameter naming and key handling

### LINKED ISSUE #794: current hset, hdel, hlen, hkeys and hexists prefix the wrong key
According to Redis' docs

https://redis.io/docs/latest/commands/hset/
https://redis.io/docs/latest/commands/hkeys/

What `django_redis.client.DefaultClient` currently calls `name` is actually the redis `key` being used, so logically, the client should call `make_key` on that, referring to the cache index, rather than on the field name in the default client.

https://github.com/jazzband/django-redis/blob/master/django_redis/client/default.py#L1145

Calling `make_key` on the field name will give you nasty surprises and end up combining hash tables with different prefixes with their "field names" prefixed.

oliverhaas:
Changed hash method parameters to align with Redis/Valkey terminology:
- Renamed 'name' → 'key' (the Redis key for the hash)
- Renamed 'key' → 'field' (the field within the hash)

**NOTE:** This is the naming convention of the Redis and Valkey docs, and unfortunately redis-py code has the naming differently, but I would go with the former, which also fits more with Django's naming.

---

Fixed make_key() application including version kwarg:
- make_key() now only applied to hash key (first parameter)
- Fields are no longer transformed with make_key()
- version kwarg was also missing from some methods

See also https://github.com/jazzband/django-redis/pull/765, and I tried to credit [2ykwang](https://github.com/jazzband/django-redis/issues?q=is%3Apr+is%3Aopen+author%3A2ykwang) in the commits, hopefully that works. Without the naming changes above their PR was a bit confusing though, so I packed this into one.

---

Let me know if I should change something here.

fixes: https://github.com/jazzband/django-redis/issues/794


tests/test_backend_hash.py
terencehonles, line 1: Please don't refactor things while you're fixing as it makes it harder to tell if you just made the tests work for broken code.

terencehonles (approved):
The typing is not correct because it is too strict (enforces Redis' types), but that is out of scope for this PR.

django_redis/client/default.py
```diff
@@ -1129,76 +1129,89 @@ def touch(
...
         """
         if client is None:
             client = self.get_client(write=True)
         nkey = self.make_key(key, version=version)
-        nvalue = self.encode(value)
-        return int(client.hset(name, nkey, nvalue))
+        return int(
+            client.hset(
+                nkey,
+                key=field,  # type: ignore[arg-type]
```
terencehonles, line 1149: There are is an issue with the Redis type hints and I will create a PR to fix those (it is a bigger change), but for now I just silenced these errors.

PR #824, opened 2026-05-30
feat: increase mypy strictness, fix typing configuration, and fix type errors

terencehonles:


pyproject.toml
```diff
@@ -76,27 +77,26 @@ django_settings_module = "tests.settings.sqlite"
 ignore_missing_settings = true
 
 [tool.mypy]
-packages = ["django_redis", "tests"]
+disallow_any_generics = false
+disallow_untyped_calls = false
+disallow_untyped_defs = false
+disallow_incomplete_defs = false
```
terencehonles, line 83: These are the strictness flags that need to remain off, but can incrementally be turned on

django_redis/client/default.py
```diff
@@ -630,8 +658,8 @@ def _incr(
                     lua = """
                     return redis.call('INCRBY', KEYS[1], ARGV[1])
                     """
-                value = client.eval(lua, 1, key, delta)
-                if value is None:
+                value: int | Literal[False] | None = client.eval(lua, 1, key, delta)
+                if value is None or value is False:
```
terencehonles, line 662: This doesn't appear to return `None` and instead was returning `False` which wasn't handled so I added this, but left the `None` check in case there is another way for the script to return `None`.

django_redis/client/sharded.py
```diff
@@ -312,25 +374,27 @@ def keys(self, search, version=None):
...
+        pattern: str,
+        version: int | None = None,
+        prefix: str | None = None,
+        client: Redis | None = None,
+        itersize: int | None = None,
+    ) -> int:
         """
         Remove all keys matching pattern.
         """
         pattern = self.make_pattern(pattern, version=version, prefix=prefix)
```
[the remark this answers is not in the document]
terencehonles: This is actually a problem on basically all of the methods, but copilot was only pointing them out one at a time and only on the methods changed. Since throwing an error on existing methods is possibly problematic I'm making it a warning and we should change the architecture a bit so that `client` is only set internally.

terencehonles:
I'm going to merge this in as it's mainly a typing change and the only user facing change should be some potential warnings and `OrderedDict` -> `dict` which unless they are using `isinstance` wouldn't have any downstream effects.

PR #830, opened 2026-06-02
release 7.0

terencehonles:


terencehonles:
@WisdomPill just a heads up I'm doing a release since it's about a year since we last released, but we can get anything that's currently in the queue in a new minor release. I went through everything and updated & merged anything that was close enough for me to pick up, but anything else will need more work to get over the fence.

WisdomPill:
@terencehonles sure, be my guest! :)

terencehonles:
:wave: oh hey, didn't think I'd get a response so quickly. Did you happen to see https://jazzband.co/news/2026/03/14/sunsetting-jazzband? I heard something about it somewhere but didn't think much about it, but just noticed that when releasing the package...

WisdomPill:
just saw it, it's the sad reality... I also did not have time recently as moved to another company which doesn't use Django hence the little time I had went down drastically... I guess we will see, it's also an option to keep checking every now and then, push this project further, perhaps with some more help and some careful usage of AI (instructions, clear way of working and roadmap), we can make this project shine again... WDYT?

